PR #1, opened 2022-04-02
Basic Rendering with piet-wgpu and tao

Demonthos:
- Lazy layout and attributes (currently only color and borders)
- Div border rendering
- Basic text rendering


univerz:
btw, do you plan to use something gtk specific from tao, that prevents you to use winit? gtk is huge and clumsy dependendy on linux.

Demonthos:
> btw, do you plan to use something gtk specific from tao, that prevents you to use winit? gtk is huge and clumsy dependendy on linux.

It would be nice to integrate with the system tray, and menu bar. Not sure if that is specific to gtk or not.

PR #2, opened 2022-06-15
fix child layout

Demonthos:
Taffy outputs layout relative to the parent. Accumulate the offsets to calculate the current position.


jkelleyrtp:
Looks better, awesome!

I plan to add text sizing too, and maybe try to integrate some more advanced text rendering techniques.

I'd love to review the native-core crate at some point. I saw you ran into some issues with GATs so it seems pretty advanced. I'm personally interested in writing custom elements using shaders as extensions for https://github.com/jkelleyrtp/sciviz and I haven't dug too deeply on how we would architect blitz to be extendable with custom elements.

PR #10, opened 2022-12-25
update to sync with master branch of Dioxus

Demonthos:
Makes blitz work with updated version of Dioxus, lightningcss, and taffy.


jkelleyrtp:
Nice!

PR #18, opened 2023-01-01
Add patch for naga to fix macOS support

Demonthos:
This applies a patch for naga to fix macOS support (https://github.com/gfx-rs/naga/pull/2099).

This should be reverted once wgpu 15 is released.


nicoburns:
I can confirm that this fixes things for me (macOS on M1)

jkelleyrtp:
Screenshot 2023-01-02 at 3 38 51 PM

Hmm.... will see if I can pin vello I guess?

PR #19, opened 2023-01-06
add support for font-size

xxrlzzz:
image

### DIFF
--- a/examples/buttons.rs
+++ b/examples/buttons.rs
@@ -47,6 +47,7 @@ fn Button(cx: Scope<ButtonProps>) -> Element {
             text_align: "center",
             display: "flex",
             flex_direction: "column",
+            font_size: "2em",
 
             p { "tabindex: {cx.props.layer}" }
         }
--- a/src/focus.rs
+++ b/src/focus.rs
@@ -79,7 +79,7 @@ impl NodeDepState for Focus {
                 if let Some(index) = a
                     .value
                     .as_int()
-                    .or_else(|| a.value.as_text().and_then(|v| v.parse::<i32>().ok()))
+                    .or_else(|| a.value.as_text().and_then(|v| v.parse::<i64>().ok()))
                 {
                     match index.cmp(&0) {
                         Ordering::Less => FocusLevel::Unfocusable,
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -40,12 +40,12 @@ impl ChildDepState for StretchLayout {
         let mut changed = false;
         if let Some(text) = node.text() {
             let mut text_context = text_context.lock().unwrap();
-            let width = text_context.get_text_width(None, 16.0, text);
+            let font_size = text_context.get_font_size(node.node_id().0);
+            let width = text_context.get_text_width(None, font_size, text);
 
             let style = Style {
                 size: Size {
-                    height: Dimension::Points(16.0),
-
+                    height: Dimension::Points(font_size),
                     width: Dimension::Points(width as f32),
                 },
                 ..Default::default()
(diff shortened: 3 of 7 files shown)


Demonthos:
This looks great, but there are some issues with native-core that need to be resolved before this can be finished.

The layout depends on the font size. When the font size changes, the layout needs to be updated to reflect that. (Try making the font size depend on if the button is focused in the buttons example and see that the text stops being centered)

The native-core system handles this automatically. Unfortunately, there is currently no good way of representing a state like the layout that is dependent based on both parent and children states.

We could probably patch this into the current system by resolving what states are dirty on a node instead of tree level, but it would requiring splitting the layout state in a way that would make the code less readable.

Once https://github.com/DioxusLabs/dioxus/pull/730 is finished it should be much easier to represent this kind of state.

xxrlzzz:
Thanks for your reply, I see the problem is the lack of mechanism that state dependency for both parent and children from `native-core`,  so their is no easy way to relayout child when parent's `font-size` got change.
As you say, this patch may require a `native-core` upgrade to keep forward.

Demonthos:
It should be possible to integrate text size with layout now that #22 has been merged

xxrlzzz:
It seems `update` doesn't get invoked when state has `ParentDependencies` and parent state change.
Since that the case that parent layout depending children font-size don't take effect. 
You can reproduce this by update parents `ForgroundColor`, there is my modify of examples/buttons.rs

```rust

fn app(cx: Scope) -> Element {
    let count = use_state(cx, || 1);
    let current_count = 10;
    let font_size = if *count.get() > 0 { "32px" } else { "16px" };
    let color = if *count.get() > 0 { "green" } else { "red" };
    cx.render(rsx! {
        div { display: "flex", flex_direction: "column", width: "100%", height: "100%",
            div {
                display: "flex",
                flex_direction: "row",
                justify_content: "center",
                align_items: "center",
                text_align: "center",
                width: "100%",
                height: "10%",
                background_color: "green",
                tabindex: "0",
                onmouseup: |_| {
                    count.modify(|c|if  *c > 0 {0} else {1} );
                },
                "grid: {current_count}x{current_count} = {current_count*current_count} tiles - Click to add more"
            }
            div {
                display: "flex",
                flex_direction: "column",
                justify_content: "center",
                align_items: "center",
                text_align: "center",
                width: "100%",
                height: "90%",
                font_size: "{font_size}",
                color: "{color}",
                (0..current_count).map(|y|
                    rsx! {
                        div { display: "flex", flex_direction: "row", width: "100%", height: "100%",
                            (0..current_count).map(|x| {
                                if (x + y) % 2 == 0 {
                                    rsx! {
                                        div {
                                            border_width: "0px",
                                            width: "100%",
                                            height: "100%",
                                            background_color: "rgb(100, 100, 100)"
                                        }
                                    }
                                } else {
                                    rsx! {
                                        Button {
                                            color_offset: x * y,
                                            layer: ((x + y) % 3) as u16
                                        }
                                    }
                                }
                            })
                        }
                    }
                )
            }
        }
    })
}

```

If there is any my misunderstood, please let me know.

Demonthos:
https://github.com/DioxusLabs/dioxus/pull/953 should fix the issue you described. The font size example 
with a similar data flow in that PR now works.

xxrlzzz:
A wired compile error
```
error[E0308]: mismatched types
  --> /Users/hanhaoshen/.cargo/git/checkouts/fount-d923d0582b4b6119/a8c0686/fello/src/meta/attributes.rs:38:31
   |
38 |             if fs_selection & FS_SELECTION_ITALIC != 0 {
   |                ------------   ^^^^^^^^^^^^^^^^^^^ expected struct `SelectionFlags`, found `u16`
   |                |
   |                expected because this is `SelectionFlags`
```

xxrlzzz:
new commit will be #29

PR #21, opened 2023-02-01
Upgrade dependencies + pin dioxus version

nicoburns:
- Upgrades dependencies
- Upgrades vello and wgpu (naga patch for macOS no longer required)
- Pins dioxus version (this was otherwise causing a compile error due to blitz requesting an older version of lightningcss which caused cargo to choose an older version of dioxus).


Demonthos (approved):
Thanks!

PR #29, opened 2023-04-19
add font size support

xxrlzzz:
Due to too much conflicts on #19 , reopen a branch and move changes to here.

image


Demonthos (approved):
It looks like the layout is working when the font size is updated now
Thank You!

PR #31, opened 2023-04-24
Invalidation

ealmloff:
Browsers generally keep track of what regions of the screen have been invalidated and re-render only those regions of the screen. This PR changes Blitz to use nodes that have been invalidated and rerender only those parts of the screen (along with any layers touching those regions)

Untitled-2023-04-18-1053


Demonthos:
It appears this may be blocked until Vello has a incremental scene API: https://github.com/linebender/vello/pull/243#issuecomment-1423391870

PR #32, opened 2023-05-16
Upgrade to Taffy 0.3

nicoburns:
Depends on: https://github.com/DioxusLabs/dioxus/pull/1030

# Objective

Support CSS Grid and gain the fixes from Taffy 0.3

## Changes made

- ~Point deps at https://github.com/DioxusLabs/dioxus/pull/1030~ PR has been merged, so this change has now been reverted.
- Upgrade Taffy version
- Add `.clone`s because `Style` is no longer `Copy`
- Add CSS Grid example

## Notes

Waiting on https://github.com/parcel-bundler/lightningcss/pull/490 in order to support `repeat()` syntax, but I don't think that ought to block this being merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,10 +11,11 @@ repository = "https://github.com/DioxusLabs/blitz"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-dioxus = { git = "https://github.com/DioxusLabs/dioxus/" }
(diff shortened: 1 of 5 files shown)


nicoburns:
Updated to work with upstream dioxus now that https://github.com/DioxusLabs/dioxus/pull/1030 has been merged (and verified to be working)

Demonthos (approved):
Thank You! 

This seems to trigger an unrelated bug (in Vello?). The outline color for the focus is red in the css_grid_buttons, but black in the buttons example:

buttons:
image
grid:
image

The PR itself looks good, so that bug does not need to block merging this.

PR #40, opened 2023-06-27
fixes build for newest versions of dependencies

nmpribeiro:
- Vello now returns `color.to_rgb()` as a Result(CssColor, ())
- Fill all match branches of FontSizeProperty abs_val


Demonthos (approved):
Thank You!

PR #41, opened 2023-07-05
blitz-core: Fix build with dioxus-native-core.

waywardmonkeys:
In May, the `layout_attributes` were moved behind a feature flag, so we need to build with that feature enabled.


Demonthos (approved):
Thank You!

PR #48, opened 2023-07-23
clippy: Fix default_constructed_unit_structs warning.

waywardmonkeys:
This warning is new in 1.71.


waywardmonkeys:
I just fixed the warning, figuring that `Config` will one day more than a unit struct and shouldn't just be removed.

Demonthos (approved):
Thank you! I must have missed this when you originally made the PR. Yes, in the future Config will not be a unit struct

PR #50, opened 2023-08-14
Remove wasm32 target dependencies.

waywardmonkeys:
Building for wasm32 doesn't work and it seems like this is intended to be a native target.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,12 +19,6 @@ blitz-core = { path = "./blitz-core" }
 tokio = { version = "1.26.0", features = ["full"] }
 keyboard-types = "0.6.2"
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
@Demonthos This one perhaps needs consideration. I'm guessing about the future directions here.

Demonthos (approved):
The only reason we would want WASM support would be for a hosted demo. It isn't a first class target of Blitz, but it would be nice as a demo of the types of scenes that Blitz can render.

We would need to remove tokio to get any WASM demo working. Instead of supporting WASM, we could just add a video to the readme with an example. This would also make it easier to implement features like notifications, file IO, etc. that need to be running outside of WASM

Even if we did support WASM, these dependancies should be in the WASM example, not the renderer

PR #54, opened 2023-10-20
Fix cssparser dependency

martijnarts:
Looks like Dioxus updated their cssparser dependency. This fixes Blitz to work with that.


ealmloff (approved):
thank you!

PR #62, opened 2024-06-08
Make the dioxus implementation interactive - just clicks for now

jkelleyrtp:
Wires up our interactivity system from dioxus into Blitz - basic buttons clicks and cursor state works now!

This is currently very rough code but can easily be picked up to make it a bit more refined.

Some methods are just meaningless/pointless/poorly organized.

https://github.com/DioxusLabs/blitz/assets/10237910/d971fc24-6626-46e2-9545-3e8121138849

### DIFF
--- a/packages/blitz/src/render.rs
+++ b/packages/blitz/src/render.rs
@@ -214,12 +214,7 @@ where
         self.mouse_pos = (x, y);
         let old_id = self.hover_node_id;
         self.hover_node_id = self.dom.as_ref().hit(x, y);
-        if old_id != self.hover_node_id {
-            true
-            // self.devtools.highlight_hover
-        } else {
-            false
-        }
+        old_id != self.hover_node_id
     }
 
     pub fn get_cursor(&self) -> Option<CursorKind> {
@@ -307,15 +302,16 @@ where
 
         // If we hit a node, then we collect the node to its parents, check for listeners, and then
         // call those listeners
-
-        self.dom.handle_event(RendererEvent {
-            name: "click".to_string(),
-            target: node_id,
-            data: EventData::Click {
-                x: self.mouse_pos.0 as f64,
-                y: self.mouse_pos.1 as f64,
-            },
-        });
+        if !self.devtools.highlight_hover {
+            self.dom.handle_event(RendererEvent {
+                name: "click".to_string(),
+                target: node_id,
+                data: EventData::Click {
+                    x: self.mouse_pos.0 as f64,
+                    y: self.mouse_pos.1 as f64,
+                },
+            });
+        }
     }
 
     pub fn print_taffy_tree(&self) {
--- a/packages/blitz/src/render.rs
+++ b/packages/blitz/src/render.rs
@@ -13,12 +13,16 @@ use crate::{
     viewport::Viewport,
 };
 use blitz_dom::{
+    events::{EventData, RendererEvent},
     node::{NodeData, TextNodeData},
     DocumentLike, Node,
 };
 use html5ever::local_name;
 use image::{imageops::FilterType, DynamicImage};
-use style::{dom::TElement, values::specified::position::HorizontalPositionKeyword};
+use style::{
+    dom::TElement,
+    values::{computed::ui::CursorKind, specified::position::HorizontalPositionKeyword},
+};
 use style::{
     properties::{style_structs::Outline, ComputedValues},
     values::{
@@ -82,8 +86,9 @@ pub struct Renderer<'s, W, Doc: DocumentLike> {
     // pub(crate) fonts: FontCache,
     pub devtools: Devtools,
 
-    hover_node_id: Option<usize>,
+    pub hover_node_id: Option<usize>,
     scroll_offset: f64,
+    mouse_pos: (f32, f32),
 }
 
 impl<'a, W, Doc: DocumentLike> Renderer<'a, W, Doc>
@@ -112,11 +117,12 @@ where
             devtools: Default::default(),
             hover_node_id: Default::default(),
             scroll_offset: 0.0,
+            mouse_pos: (0.0, 0.0),
         }
     }
 
-    pub fn poll(&mut self, cx: std::task::Context) {
-        self.dom.poll(cx);
+    pub fn poll(&mut self, cx: std::task::Context) -> bool {
+        self.dom.poll(cx)
     }
 
     pub async fn resume(&mut self, window_builder: impl FnOnce() -> (Arc<W>, Viewport)) {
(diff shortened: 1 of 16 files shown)


packages/blitz/src/render.rs
```diff
@@ -196,16 +202,53 @@ where
...
         let old_id = self.hover_node_id;
         self.hover_node_id = self.dom.as_ref().hit(x, y);
         if old_id != self.hover_node_id {
-            // println!("Hovered node: {:?}", self.hover_node_id);
-            self.devtools.highlight_hover
+            true
+            // self.devtools.highlight_hover
         } else {
             false
         }
```
nicoburns, line 222: This whole block could now be simplified to `old_id != self.hover_node_id`

packages/blitz/src/render.rs
```diff
@@ -196,16 +202,53 @@ where
...
     }
 
+    pub fn get_cursor(&self) -> Option<CursorKind> {
+        // todo: cache this on the node itself
+        let node = &self.dom.as_ref().tree()[self.hover_node_id.unwrap()];
```
nicoburns, line 227: I think we want to set this in the Stylo `ElementData`. That (+ some cache invalidation) ought to make hover styles work I think.

packages/blitz/src/render.rs
```diff
@@ -237,24 +280,42 @@ where
...
+        // call those listeners
+
+        self.dom.handle_event(RendererEvent {
+            name: "click".to_string(),
+            target: node_id,
+            data: EventData::Click {
+                x: self.mouse_pos.0 as f64,
+                y: self.mouse_pos.1 as f64,
+            },
+        });
```
nicoburns, line 318: This should probably be in an `else`. Devtools inspector being enabled usually disables the app from handling clicks (clicks highlight element in devtools instead).

packages/dioxus-blitz/src/documents/dioxus_document.rs
```diff
@@ -47,26 +52,241 @@ impl Into<Document> for DioxusDocument {
...
-                    std::task::Poll::Pending => break,
+                    std::task::Poll::Pending => return false,
                 }
             }
+
+            self.vdom.render_immediate(&mut MutationWriter {
+                doc: &mut self.inner,
+                state: &mut self.vdom_state,
+            });
+            println!("wrote mutatins to dom");
```
nicoburns, line 71: Typo mutatins

packages/dioxus-blitz/src/documents/dioxus_document.rs
```diff
@@ -47,26 +52,241 @@ impl Into<Document> for DioxusDocument {
...
+            return true;
         }
     }
+
+    fn handle_event(&mut self, event: blitz_dom::events::RendererEvent) -> bool {
+        // Collect the nodes into a chain by traversing upwards
+        // This is important so the "capture" phase can be implemented
+        let mut node = event.target;
+        let mut chain = vec![node];
```
nicoburns, line 80: Nit: Would be nice to preallocate a Vec of size 16 or similar.

nicoburns: Nit: I've been trying to use `node_id` when it's a `NodeId` and `node` when it's a `Node`.

packages/dioxus-blitz/src/documents/dioxus_document.rs
```diff
@@ -47,26 +52,241 @@ impl Into<Document> for DioxusDocument {
...
+                todo!()
+            }
+
+            fn held_buttons(&self) -> dioxus::prelude::dioxus_elements::input_data::MouseButtonSet {
+                todo!()
+            }
+        }
+        impl dioxus::html::HasMouseData for NativeClickData {
+            fn as_any(&self) -> &dyn std::any::Any {
+                todo!()
```
nicoburns, line 139: Could probably implement this ;)

packages/dioxus-blitz/src/documents/dioxus_document.rs
```diff
@@ -251,6 +471,18 @@ impl WriteMutations for MutationWriter<'_> {
...
+
+        if let Some(parent) = parent {
+            // if the text is the child of a style element, we want to put the style into the stylesheet cache
+            let parent = self.doc.get_node(parent).unwrap();
+            if let NodeData::Element(ref element) = parent.raw_dom_data {
+                if element.name.local.as_ref() == "style" {
+                    self.doc.add_stylesheet(value);
+                }
+            }
+        }
```
nicoburns, line 485: To a be a little more robust, I was going to suggest push style tags to a queue when they are encountered. Then processing once the mutations list has been processed. We'll also need to deal with edits to existing stylesheets. And making sure multiple stylesheets are added in the right order ("document order").

This will probably do for now though...

packages/dioxus-blitz/src/documents/dioxus_document.rs
```diff
@@ -338,14 +570,48 @@ impl WriteMutations for MutationWriter<'_> {
...
+        };
+
+        println!("set_node_text id:{} text:{}", id.0, value);
+
+        if let Some(parent) = parent {
+            // if the text is the child of a style element, we want to put the style into the stylesheet cache
+            let parent = self.doc.get_node(parent).unwrap();
+            if let NodeData::Element(ref element) = parent.raw_dom_data {
+                if element.name.local.as_ref() == "style" {
+                    self.doc.add_stylesheet(value);
```
nicoburns, line 591: This probably also needs to remove the existing stylesheet (if any). Otherwise mutations to stylesheets are going to end up creating duplicates

packages/dioxus-blitz/src/window.rs
```diff
@@ -59,6 +59,19 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
...
+            node.primary_styles().map(|f| f.clone_cursor().keyword)
+        };
+
+        if let Some(cursor) = cursor {
+            // self.set_cursor(cursor);
+        } else {
+            // self.set_cursor(style::values::computed::ui::CursorKind::Auto);
+        }
+    }
+
```
nicoburns, line 74: This isn't doing anything? (I also don't like this unwrap)

packages/dioxus-blitz/src/window.rs
```diff
@@ -152,10 +167,78 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
                 // modifiers,
                 ..
             } => {
-                let tao::dpi::LogicalPosition::<f32> { x, y } = position.to_logical(2.0);
-                if self.renderer.mouse_move(x, y) {
-                    self.request_redraw();
+                let changed = if let RenderState::Active(state) = &self.renderer.render_state {
+                    // todo: this is a hardcode of the hidpi scale, we should get this from the window
```
nicoburns, line 171: Looks like you might be doing this now?

packages/dioxus-blitz/src/window.rs
```diff
@@ -152,10 +167,78 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
...
+                if changed {
+                    let cursor = self.renderer.get_cursor();
+                    dbg!(cursor);
+
+                    if let Some(cursor) = cursor {
+                        use style::values::computed::ui::CursorKind;
+                        use tao::window::CursorIcon as TaoCursor;
+                        let tao_cursor = match cursor {
+                            CursorKind::None => todo!("set the cursor to none"),
+                            CursorKind::Default => TaoCursor::Default,
```
nicoburns, line 188: Maybe a `_ => cursor` catch-all for the cases where no mapping is happening. Normally I like to write out all the variants, but it doesn't seem important here...

packages/dom/src/layout.rs
```diff
@@ -65,9 +65,13 @@ impl LayoutPartialTree for Document {
...
 
+            // this is because of hidpi affecting the layout - we need to propagate this down
+            // the 2.0 is the dpi and the .2 is a bit extra since our measure function is a bit off
+            const FUDGE_FACTOR: f32 = 2.2;
+
             let font_metrics = FontMetrics {
-                char_width: 8.0,
-                char_height: 16.0,
+                char_width: 8.0 * FUDGE_FACTOR,
+                char_height: 16.0 * FUDGE_FACTOR,
```
nicoburns, line 74: Fine since parley is coming soon. The better fudge would be a fixed version of https://github.com/DioxusLabs/blitz/commit/024c26c0ca59945e2469c5c4a863f06b5621141f (which I had to revert because it was breaking the google example. We're currently hardcoding font size for layout but computing for rendering!

nicoburns (approved):

PR #64, opened 2024-06-13
Fix building and few crashes on Linux

DavidHusicka:
Project could not be compiled on Linux due to missing import. This import was added.

Crashes when resizing were also common so they were solved with an early return.

Lastly, the `get_cursor` function sometimes unwraped a `None` variant on an `Option` so the `unwrap` was replaced with an error propagation syntax.


nicoburns (approved):
This looks great - thanks! Hopefully we can look at getting some CI checks in soon, to make sure that we can always build on all platforms.

PR #66, opened 2024-06-15
Add basic support for underline and strikethrough decorations

kokoISnoTarget:
lines


packages/blitz/src/render.rs
```diff
@@ -788,6 +788,58 @@ impl ElementCx<'_> {
                                 }
                             }),
                         );
+
+                    if let Some(underline) = &style.underline {
+                        let mut draw_underline = |x: f32, y: f32, w: f32| {
+                            let thickness = 1.0 / 18.0;
+                            let offset = 1.0 / 9.0;
```
nicoburns, line 795: The thickness and offset should be come from the `style` and fall back to a value from `run_metrics`  (`glyph_run.run().metrics()`) rather than being hardcoded. See https://github.com/linebender/xilem/blob/main/masonry/src/text_helpers.rs#L126. I would suggest using `Option::unwrap_or` rather than the `match` statement in the linked code.

packages/blitz/src/render.rs
```diff
@@ -788,6 +788,58 @@ impl ElementCx<'_> {
...
+                            let path = Rect::new(
+                                x as f64,
+                                y as f64 + font_size as f64 * offset,
+                                (x + w) as f64,
+                                y as f64 + font_size as f64 * (thickness + offset),
+                            );
+
+                            scene.fill(
```
nicoburns, line 803: Consider using `peniko::Line` with `scene.stroke` rather than `kurbo::Rect` with `scene.fill`. It simplifies this code and I think a stroke is more appropriate than a fill for drawing a simple line like this. See https://github.com/linebender/xilem/blob/main/masonry/src/text_helpers.rs#L139

packages/blitz/src/render.rs
```diff
@@ -788,6 +788,58 @@ impl ElementCx<'_> {
                                 }
                             }),
                         );
+
+                    if let Some(underline) = &style.underline {
+                        let mut draw_underline = |x: f32, y: f32, w: f32| {
```
nicoburns, line 793: Perhaps this function (with a couple of extra params) should be shared between underline and strikethrough?

packages/blitz/src/render.rs
```diff
@@ -788,6 +788,58 @@ impl ElementCx<'_> {
...
+                            );
+                        };
+                        draw_underline(
+                            glyph_run.offset(),
+                            glyph_run.baseline(),
+                            glyph_run.advance(),
+                        )
+                        // todo: intercept line when crossing an descending character like "gqy"
+                    }
+                    if let Some(strikethrough) = &style.strikethrough {
```
nicoburns, line 818: Same comments as for underline.

packages/dom/src/stylo_to_parley.rs
```diff
@@ -98,6 +98,18 @@ pub(crate) fn style(style: &stylo::ComputedValues) -> parley::TextStyle<'static,
...
 
+    let decoration_brush = style.get_text()
+            .text_decoration_color
+            .as_absolute()
+            .map(|color| {
+                let [r, g, b, a] = color
+                    .to_color_space(style::color::ColorSpace::Srgb)
+                    .raw_components()
+                    .map(|f| (f * 255.0) as u8);
+                let color = peniko::Color { r, g, b, a };
```
nicoburns, line 109: Could this color conversion logic be pulled out into a function please? (seeing as it is now duplicated)

nicoburns (changes requested):
Thanks for the PR. This broadly looks good, but I have requested a couple of changes around the source of the offset/thickness of the line, and made some suggestions around improving code quality.

Also: feel free to tick off `underline/strikethrough` in the README.

packages/dom/src/stylo_to_parley.rs
kokoISnoTarget: I've implemented a helper, but it's basically the same as in [/blitz/src/utils.rs](https://github.com/DioxusLabs/blitz/blob/133608dab382b327843f1f1c11753e0cfabc089d/packages/blitz/src/util.rs#L35C2-L57C6) just in dom.

nicoburns: Yeah, I'm aware of the duplication but we can't use the code from `blitz` in `blitz-dom`. We might be able to make the dep go the other way. But it's also a small enough amount of code that maybe we can just duplicate it (I'm not anticipating there being a huge amount more code that needs to be duplicated as there is mostly quite a neat layout vs. presentation split - but color happens to be needed in both).

nicoburns (approved):

nicoburns:
Ah. I should have tested this before merging. Something is not quite right...

Screenshot 2024-06-18 at 11 42 33

nicoburns:
Fixed! It just needed to be `transform` rather than `self.transform`.

PR #67, opened 2024-06-18
Hot reloading support

matthunz:
Connects `blitz` to the `dioxus-hot-reload` crate by propagating `tao` events to the `DioxusDocument`.

Blitz examples still don't seem to work correctly, I think it has something to do with the virtual manifest?


Cargo.toml
```diff
@@ -32,19 +32,14 @@ edition = "2021"
...
 opt-level = 2
 
 # Need to force specific versions of these dependencies
 [dependencies]
-# webrender = "0.61.0"
-euclid = { version = "0.22", features = ["serde"] }
-# mozbuild = "0.1.0"
-
-[dev-dependencies]
 blitz = { path = "./packages/blitz" }
```
nicoburns, line 43: Could this change be causing the examples not to work correctly?

packages/dioxus-blitz/Cargo.toml
```diff
@@ -3,13 +3,17 @@ name = "dioxus-blitz"
...
+hot-reload = []
+default = []
 
 [dependencies]
 winit = { version = "0.30.2", features = ["rwh_06"] }
 muda = { version = "0.11.5", features = ["serde"] }
 tokio = { workspace = true, features = ["full"] }
 dioxus = { workspace = true }
+dioxus-cli-config = { git = "https://github.com/dioxuslabs/dioxus", rev = "a3aa6ae771a2d0a4d8cb6055c41efc0193b817ef"}
+dioxus-hot-reload = { git = "https://github.com/dioxuslabs/dioxus", rev = "a3aa6ae771a2d0a4d8cb6055c41efc0193b817ef"}
```
nicoburns, line 16: These should be optional, based on the `hot-reload` feature.

packages/dom/src/document.rs
```diff
@@ -217,7 +222,9 @@ impl Document {
 
         let node = &self.nodes[node_id];
         let node_child_idx = node.child_idx;
-        let parent_id = node.parent.unwrap();
+
+        // Get this node's parent, or the root node if it has none.
+        let parent_id = node.parent.unwrap_or_default();
```
nicoburns, line 227: This doesn't seem right. Only the root node should have no parent. And we wouldn't want to insert a child of the root node instead of a sibling. What is the use case here? Replacing the root node?

nicoburns (changes requested):

ealmloff (approved):
The hot reloading part of this looks good! It is very similar to how desktop does things which is nice

packages/dom/src/document.rs
nicoburns: @matthunz This is the only thing blocking this PR as far as I'm concerned. I'd like to understand why this is required. It seems wrong that `WriteMutations::insert_nodes_after` would be called on a node without a parent.  @ealmloff Any thoughts?

ealmloff: I don't think this should be necessary. The root node should be `ElementId(0)` and every node except that node should have a parent in dioxus mutations. Dioxus will never insert a node after or before `ElementId(0)`

PR #75, opened 2024-06-20
Fix cross-platform mudabar, wire up better CI

jkelleyrtp:
Needed to fix some issues regarding muda - looks like we never implemented some of it.

Had to allow deprecated on the closure approach for winit.

Stole our CI for multiplatform from dioxus

(todo) override opt-level for CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,42 +86,32 @@ jobs:
           - {
               target: x86_64-pc-windows-msvc,
               os: windows-latest,
-              toolchain: "1.75.0",
               cross: false,
               command: "test",
               args: "--all --tests",
+              setup: ""
             }
           - {
               target: x86_64-apple-darwin,
               os: macos-latest,
-              toolchain: "1.75.0",
               cross: false,
               command: "test",
               args: "--all --tests",
+              setup: ""
             }
           - {
               target: x86_64-unknown-linux-gnu,
               os: ubuntu-latest,
-              toolchain: "1.75.0",
               cross: false,
               command: "test",
               args: "--all --tests",
-            }
-          - {
-              target: aarch64-apple-ios,
-              os: macos-latest,
-              toolchain: "1.75.0",
-              cross: false,
-              command: "build",
-              args: "--package dioxus-mobile",
-            }
-          - {
-              target: aarch64-linux-android,
-              os: ubuntu-latest,
-              toolchain: "1.75.0",
-              cross: true,
-              command: "build",
-              args: "--package dioxus-mobile",
+              setup: "sudo apt-get update; sudo apt-get install --no-install-recommends \
+                libasound2-dev \
+                libatk1.0-dev \
+                libgtk-3-dev \
+                libudev-dev \
+                libpango1.0-dev \
+                libxdo-dev"
             }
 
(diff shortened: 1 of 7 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
nicoburns, line 1: Committing this seems like a bad idea as it means that people won't be able to set their own settings. Maybe a "template" file with a different name?

packages/dioxus-blitz/src/window.rs
```diff
@@ -290,26 +290,14 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
...
-            // {
-            //     menu_bar.init_for_nsapp();
-            //     build_menu().set_as_windows_menu_for_nsapp();
-            // }
+            init_menu_bar(&menu_bar, &window);
+
+            // for now, forget the menu_bar so it doesn't get dropped
+            // there's a bug in muda that causes this to segfault
+            // todo: we should just store this somewhere
+            std::mem::forget(menu_bar);
```
nicoburns, line 300: Could this just be a field of `View`?

.github/workflows/ci.yml
```diff
@@ -69,3 +73,84 @@ jobs:
...
+              command: "test",
+              args: "--all --tests",
+            }
+          - {
+              target: aarch64-apple-ios,
+              os: macos-latest,
+              toolchain: "1.75.0",
+              cross: false,
+              command: "build",
+              args: "--package dioxus-mobile",
```
nicoburns, line 116: Building `dioxus-mobile` in blitz CI probably doesn't make sense?

.github/workflows/ci.yml
```diff
@@ -69,3 +73,84 @@ jobs:
...
+    if: github.event.pull_request.draft == false
+    env:
+      RUST_CARGO_COMMAND: ${{ matrix.platform.cross == true && 'cross' || 'cargo' }}
+    strategy:
+      matrix:
+        platform:
+          - {
+              target: x86_64-pc-windows-msvc,
+              os: windows-latest,
+              toolchain: "1.75.0",
```
nicoburns, line 89: Why `1.75`?

nicoburns: Also, does `toolchain` need to be specified per-platform. Could just make global?

nicoburns (changes requested):

packages/dioxus-blitz/src/window.rs
```diff
@@ -23,7 +25,6 @@ use winit::platform::unix::WindowExtUnix;
 use winit::platform::windows::WindowExtWindows;
```
nicoburns, line 25: Looks like winit has `wayland` and `x11` platforms rather than `unix`.

.github/workflows/ci.yml
nicoburns: We could just skip on android/ios for the time being

.vscode/settings.json
matthunz: Totally agree here 👍 maybe `.vscode.template` or similar? This was really helpful for me (I didn't know about `allTargets` 👀) so other people might find that nice too

Issue #14, opened 2022-12-28
Integration with Accesskit


ealmloff:
Blitz should integrate with [accesskit](https://crates.io/crates/accesskit_winit) to provide integration with platform accessablity apis.

PR #78, opened 2024-06-20
AccessKit integration

matthunz:
Starts work on AccessKit integration (https://github.com/DioxusLabs/blitz/issues/14)
 - [x] Connect `accesskit-winit` adapter
 - [x] Add `Dom::visit` to traverse the dom top-down
 - [x] Build initial AccessKit tree
 - [x] Send tree changes from Dioxus updates to adapter

I think it's best to wait for more interactivity and focus system design before processing AccessKit events with the DOM 🤔   totally open to trying to add that here though


packages/dioxus-blitz/src/window.rs
```diff
@@ -308,6 +398,47 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
         self.waker = None;
         self.renderer.suspend();
     }
+
+    #[cfg(feature = "accesskit")]
+    fn build_accessibility_tree(&mut self) {
```
nicoburns, line 403: Can we move this function (and probably `build_node` above too) into their own `accesskit` module?

packages/dom/src/document.rs
```diff
@@ -522,6 +538,23 @@ impl Document {
...
+        while let Some(node_key) = stack.pop_back() {
+            let node = &self.nodes[node_key];
+            visit(node_key, node);
+
+            for &child_key in &node.children {
+                stack.push_front(child_key);
+            }
+        }
+    }
 }
```
nicoburns, line 558: Not sure about this (but not totally against it either). You could potentially use the pattern where you `mem::take` the children array before recursing into the children, and then put it back afterwards. That's not exactly ideal either though. Perhaps this pattern is better?

packages/dom/src/document.rs
```diff
@@ -87,6 +92,8 @@ pub struct Document {
     pub(crate) layout_ctx: parley::LayoutContext<TextBrush>,
 
     pub(crate) hover_node_id: Option<usize>,
+
+    pub changed: HashSet<usize>,
```
nicoburns, line 96: Why is this a `HashSet` rather than a `Vec`? It never seems to be randomly queried, only ever iterated over?

packages/dioxus-blitz/src/window.rs
```diff
@@ -45,7 +97,20 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
...
                 let cx = std::task::Context::from_waker(waker);
-                self.renderer.poll(cx)
+                if self.renderer.poll(cx) {
+                    #[cfg(feature = "accesskit")]
+                    {
+                        // TODO send fine grained accessibility tree updates.
+                        let changed = mem::take(&mut self.renderer.dom.as_mut().changed);
+                        if !changed.is_empty() {
+                            self.build_accessibility_tree()
+                        }
```
nicoburns, line 107: Seems like it wouldn't be too hard to send a `TreeUpdate` with just changed nodes here?

packages/dioxus-blitz/src/window.rs
```diff
@@ -15,6 +16,57 @@ use winit::event::{ElementState, MouseButton};
...
+            };
+
+            node_builder.set_role(role);
+            node_builder.set_name(name);
+        } else if node.is_text_node() {
+            node_builder.set_role(accesskit::Role::StaticText);
+            node_builder.set_name(node.text_content());
+        }
+
+        let id = accesskit::NodeId(self.next_id);
```
nicoburns, line 60: Oh... we get to choose the AccessKit ids... Is there any reason not to use the dom's NodeId directly?

packages/dioxus-blitz/src/window.rs
```diff
@@ -268,10 +333,23 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
...
+    #[cfg(feature = "accesskit")]
+    pub fn handle_accessibility_event(&mut self, event: &accesskit_winit::WindowEvent) {
+        match event {
+            accesskit_winit::WindowEvent::InitialTreeRequested => self.build_accessibility_tree(),
+            accesskit_winit::WindowEvent::AccessibilityDeactivated => {
+                // TODO
+            }
+            accesskit_winit::WindowEvent::ActionRequested(_req) => {
+                // TODO
+            }
```
nicoburns, line 345: Might it be worth at least implementing "click" here (which of course is the only event we support generally. I think the code could just be copied from the regular click handler (skipping the bit which determines which node to generate the event for as we should be given a node id directly)

Cargo.toml
```diff
@@ -32,19 +32,14 @@ edition = "2021"
...
 [profile.dev.package."*"]
 opt-level = 2
 
 # Need to force specific versions of these dependencies
 [dependencies]
-# webrender = "0.61.0"
-euclid = { version = "0.22", features = ["serde"] }
-# mozbuild = "0.1.0"
-
-[dev-dependencies]
```
nicoburns, line 47: What's the reason for moving all of these deps from "dev-dependencies" to "dependencies"?

nicoburns:
@matthunz Have you tried actually using this with a screenreader?

nicoburns (changes requested):

matthunz:
> @matthunz Have you tried actually using this with a screenreader?

Yeah it looks to be reading the elements OK so far, I posted this video in the Dioxus discord a little while back if you missed it

https://github.com/DioxusLabs/blitz/assets/9288430/62c7f072-2d76-46bc-910a-b169b287a631

packages/dioxus-blitz/src/window.rs
```diff
@@ -282,10 +360,22 @@ impl<'a, Doc: DocumentLike> View<'a, Doc> {
...
-            ));
+            );
+            self.state = Some(State {
+                #[cfg(feature = "accesskit")]
+                adapter: accesskit_winit::Adapter::with_event_loop_proxy(&window, proxy.clone()),
+                #[cfg(feature = "accesskit")]
+                node_ids: Default::default(),
+                #[cfg(feature = "accesskit")]
+                id_to_node_ids: Default::default(),
+                #[cfg(feature = "accesskit")]
```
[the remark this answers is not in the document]
matthunz: For sure, I like this a lot, those `cfg` flags are super ugly
https://github.com/DioxusLabs/blitz/blob/0ce74dd7c262c892e512f146e32c55a4af2005a6/packages/dioxus-blitz/src/accessibility.rs#L7

packages/dioxus-blitz/src/window.rs
```diff
@@ -15,6 +16,57 @@ use winit::event::{ElementState, MouseButton};
...
+        &mut self,
+        node_id: usize,
+        node: &Node,
+    ) -> (accesskit::NodeId, accesskit::NodeBuilder) {
+        let mut node_builder = accesskit::NodeBuilder::default();
+        if let Some(element_data) = node.element_data() {
+            let name = element_data.name.local.to_string();
+            let role = match &*name {
+                "button" => accesskit::Role::Button,
+                "div" | "section" => accesskit::Role::Group,
```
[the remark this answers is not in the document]
matthunz: Oh thanks for the tips, I added those and some others I could think of here so far https://github.com/DioxusLabs/blitz/blob/0ce74dd7c262c892e512f146e32c55a4af2005a6/packages/dioxus-blitz/src/accessibility.rs#L62

packages/dioxus-blitz/src/window.rs
```diff
@@ -15,6 +16,57 @@ use winit::event::{ElementState, MouseButton};
...
 
+struct State {
+    /// Accessibility adapter for `accesskit`.
+    #[cfg(feature = "accesskit")]
+    adapter: accesskit_winit::Adapter,
+
+    #[cfg(feature = "accesskit")]
+    node_ids: std::collections::HashMap<accesskit::NodeId, usize>,
+
+    #[cfg(feature = "accesskit")]
```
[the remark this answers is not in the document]
matthunz: This makes sense to me but would we need to move `accesskit` to Blitz core for that? For now I removed the field until we have support for accessibility events (that doesn't seem clear to me to implement yet unless you have any ideas 🤔 )

packages/dom/src/document.rs
matthunz: Hmm I'm not sure I understand, so we'd be clearing the `children` field while visiting a node's children? I can see that preventing cycles if that's what you're thinking

packages/dom/src/document.rs
matthunz: My plan was to eventually have this be used for the fine-grained accessibility tree updates but I don't see clear path for that yet 🤔 we could totally switch this to a `bool` for now if we want to stick with the coarse updates for now.

Personally I think we can get away with coarse-grained for a little while before apps in Blitz start getting overly complex

packages/dioxus-blitz/src/window.rs
matthunz: I think you're right this is how we should handle this, I was just having some trouble getting the hierarchy working. The issue I had was creating IDs for the accesskit elements so maybe re-using the Blitz element IDs here like you mentioned could help?

packages/dioxus-blitz/src/window.rs
matthunz: If we are able to map 1-1 from nodes (and text nodes) I'd say we should totally do this 👀

packages/dioxus-blitz/src/accessibility.rs
```diff
@@ -0,0 +1,92 @@
...
+                    match ty {
+                        "number" => Role::NumberInput,
+                        _ => Role::TextInput,
+                    }
+                }
+                _ => Role::Unknown,
+            };
+
+            node_builder.set_role(role);
+            node_builder.set_name(name);
```
DataTriny, line 81: In this scope, the variable `name` seem to always contain the "HTML element type" which is not something the user should see. You could call `NodeBuilder::set_html_tag` (although we currently don't expose this property to assistive technologies) but you should not use it for the AccessKit node name.

DataTriny (changes requested):

DataTriny:
Hello @matthunz, AccessKit dev here.

I've tried a modified version of the accessibility example to add more kind of nodes and my screen reader is already able to pick them up.
One thing you will have to tackle early on is the `labelled_by` relationship. I tried an `h1` and a `button` and on both cases the text that my screen reader attaches to the node is the HTML element node. I assume that the actual textual content is a child node of the element, so you will have to put the text node's ID in the `labelled_by` list property on the parent container.

matthunz:
Hey @DataTriny thanks for your insight!

I switched `NodeBuilder::set_name` to `NodeBuilder::set_html_tag` in https://github.com/DioxusLabs/blitz/pull/78/commits/9ae0fb7999b827edccd6a49ed3ed069f1bdc3514 . If there is a plan to send this to the platform eventually I'd say we could keep it here until then :thinking: 

I also tried added some basic `labelled_by` relationships in https://github.com/DioxusLabs/blitz/pull/78/commits/a882185736facb914e9d104ebc372aa39f5984a3 . I'm hoping concatenating child text node ids into the parent elements `labelled_by` field should give the right result.

I'd also really like to connect to `aria-labelled-by` at some point :eyes:

packages/dioxus-blitz/src/window.rs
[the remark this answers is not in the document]
nicoburns: > but would we need to move accesskit to Blitz core for that?

I don't think so? Node IDs are just usize atm, and are accessible from outside of the crate. We'll might want to switch to a newtype at some point, but I'm sure we'll make conversions to plain types available.

packages/dom/src/document.rs
nicoburns: Ah yes, `HashSet` for deduplicating insertions. Makes sense.

nicoburns (approved):
Happy to accept this as an initial implementation and iterate :)

PR #79, opened 2024-06-20
Use opt-level = 0 in CI

### LINKED ISSUE #71: Speed up CI by using opt-level 0 for dependencies
Blitz specifies a default opt-level of 2 for dependencies. This causes CI to be much slower than it needs to be. We should override this setting and use opt-level 0 in CI.

nicoburns:
Fixes #71


nicoburns:
Windows is still kinda slow but this is definitely better.

PR #86, opened 2024-06-25
Initial Android support

matthunz:
Provides the initial working support for Android. Interactivity is blocked on touch and gesture support, but HTML can be displayed and seems to work across app launches.

This also adds `docsrs` configuration to provide information about android-only functions


packages/dioxus-blitz/src/lib.rs
```diff
@@ -93,17 +136,25 @@ fn launch_with_window<Doc: DocumentLike + 'static>(window: View<'static, Doc>) {
...
     let proxy = event_loop.create_proxy();
 
     // Multiwindow ftw
     let mut windows: HashMap<WindowId, window::View<'_, Doc>> = HashMap::new();
     let mut pending_windows = Vec::new();
 
     pending_windows.push(window);
-    let menu_channel = MenuEvent::receiver();
+
+    #[cfg(feature = "muda")]
```
nicoburns, line 156: I reckon this feature should probably be called "menu" "menubar".

nicoburns (commented):
Mostly looks good to me. I wonder whether we need to gate the menu stuff on "not android platform" too though? Otherwise you'd have to build with different feature flags on different platforms. Maybe that's fine?

packages/dioxus-blitz/src/lib.rs
matthunz: Totally agree 👍 I went with the name `menu` in 823f309a566f45a231dc237b32ffa82349d419a0 but definitely open to `menubar` as well 

I also made the feature flag default, but not usable from Android/IOS, which I think could be easier for end users

PR #107, opened 2024-07-29
Webfonts WIP

nicoburns:
This one isn't working yet either:
- Servo.org also requires `@import` support for it's webfonts
- Dioxuslabs.com loads the font but REALLY slowly (due to sync IO I think). We're going to need a proper tokio-based network thread before we land this.
- The fonts still don't actually take affect. We may need to improve support in Fontique for this.


nicoburns:
Implemented in #123 and #138

PR #113, opened 2024-08-07
Reuse anonymous blocks for complex inline layouts

DavidHusicka:
This gets rid of a memory leak where the anonymous block would be created in a loop and never cleared.


packages/dom/src/layout/construct.rs
```diff
@@ -235,6 +235,13 @@ fn collect_complex_layout_children(
         else if needs_wrap(child_node_kind, display_outside) {
             use style::selector_parser::PseudoElement;
 
+            // TODO: once let chaining lends in stable, make nicer
```
nerdachse, line 238: Lends -> lands

nicoburns:
This has now been implemented

PR #115, opened 2024-08-12
Fix a memory leak with Parley fonts

DavidHusicka:
Using a self-referencing struct and bunch of unsafe, we are able to extend the lifetime of the font names and their array without leaking them. These are necessary for a TextStyle which is part of Parley. This creates a shim called "OwnedTextStyle" that owns the data.

### DIFF
--- a/packages/dom/src/layout/construct.rs
+++ b/packages/dom/src/layout/construct.rs
@@ -281,11 +281,12 @@ fn collect_complex_layout_children(
 
 fn create_text_editor(doc: &mut Document, input_element_id: usize, is_multiline: bool) {
     let node = &mut doc.nodes[input_element_id];
-    let parley_style = node
+    let mut parley_style = node
         .primary_styles()
         .as_ref()
         .map(|s| stylo_to_parley::style(s))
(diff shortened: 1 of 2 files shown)


nicoburns:
> Using a self-referencing struct and bunch of unsafe, we are able to extend the lifetime of the font names and their array without leaking them. These are necessary for a TextStyle which is part of Parley. This creates a shim called "OwnedTextStyle" that owns the data.

Hmm... all the unsafe code with tricky lifetimes doesn't seem ideal. How about instead of a self-referencing struct, the names are stored in a document-global HashSet (or similar- perhaps a HashSet wouldn't work). There are unlikely to be many unique font names used within a single Document so this shouldn't use too much memory.

Or even a short-lived datastructure (that is passed in to the stylo_to_parley function so it is high enough up the stack for the lifetimes to work. I believe the FamilyNames style only needs to live until it is used (the reference is not stored).

nicoburns:
This has now been fixed as part of the upgrade to Parley 0.2 which allows owned styles to be used.

PR #116, opened 2024-08-12
Pass focus to accesskit and simplify node ids

matthunz:
- Unifies AccessKit IDs with `Document` element IDs
 - Adds the `Document::focus` method and calls it to send the currently focused ID to AccessKit
 - Process window events with AccessKit
 - TODO fix ID errors in interactive examples

### DIFF
--- a/packages/dom/src/document.rs
+++ b/packages/dom/src/document.rs
@@ -67,6 +67,7 @@ pub trait DocumentLike: AsRef<Document> + AsMut<Document> + Into<Document> + 'st
     }
 }
 
+// TODO (Matt): change document IDs to be u64 for accesskit compatibility.
 pub struct Document {
     /// A bump-backed tree
     ///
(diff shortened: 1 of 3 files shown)


packages/dom/src/document.rs
```diff
@@ -578,6 +579,10 @@ impl Document {
         }
     }
 
+    pub fn focus(&self) -> Option<usize> {
```
nicoburns, line 582: I think you should probably use the existing `get_focussed_node_id` for this.

packages/dom/src/document.rs
```diff
@@ -67,6 +67,7 @@ pub trait DocumentLike: AsRef<Document> + AsMut<Document> + Into<Document> + 'st
     }
 }
 
+// TODO (Matt): change document IDs to be u64 for accesskit compatibility.
```
nicoburns, line 70: I suspect we won't want to do this. `usize` is convenient for indexing. Probably at some point we'll want our own `NodeId` newtype. Then we can implement indexing and from/into conversions for it, and gain some type safety in the process.

nicoburns (changes requested):
Mostly looks good. A couple of suggestions!

PR #120, opened 2024-08-15
Add a root html element in Dioxus integration

### LINKED ISSUE #81: Blitz (dioxus runner) crashes if there is more than one root node
Perhaps we should consider creating the `<html>` node automatically? Or alternatively making the code tolerant of there not being one.

cfraz89:
Fixes #81 - Currently, dioxus apps without a single root element crash when rendered in blitz. This fixes that by creating a virtual `<html>` root element. 

I experimented with a couple of other approaches as well, such as:

- treating Document nodes as an element as well - a bit of an impedance mismatch since stylo makes a distinction between documents and elements
- abolishing the concept of a root element and rendering all child elements of the root node - mismatch since taffy wants to draw a root element.

So this way seemed to be the best way to make it work out. With this fix you should find that the dioxus examples such as `bare_style` and `tailwind` no longer crash, and render correctly as far as they can.

### DIFF
--- a/packages/dioxus-blitz/src/documents/dioxus_document.rs
+++ b/packages/dioxus-blitz/src/documents/dioxus_document.rs
@@ -138,15 +138,15 @@ impl DioxusDocument {
         let viewport = Viewport::new(0, 0, 1.0);
         let mut doc = Document::new(viewport);
 
-        // Create a virtual "dioxus:document" element to act as the root element, as we won't necessarily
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
We'll want to think this through properly at some point (I suspect we may want to align Blitz so that the same code can easily be used across Blitz and other other dioxus platforms). But I reckon this is better than we have.

PR #121, opened 2024-08-17
Non-root  scrolling

cfraz89:
Implements scrolling on mouse-wheel events, of any node which is hovered. When a node has hit its scroll boundary, scrolling bubbles up to the node's parent, all the way up to the root element. The root (html) element is sized to the viewport (is this technically the right thing to do?) and therefore will automatically handle root-level scrolling.

Adds an example `scrolling` which demos scrolling in the middle element. Scrolling the root element can be tested by resizing the window smaller than the content and scrolling, otherwise the tailwind example has a large content area and should be scrollable too.

Todo: Touch scrolling, flinging with momentum (even with mouse wheel? chromium does this).

### DIFF
--- a/packages/blitz/src/renderer/render.rs
+++ b/packages/blitz/src/renderer/render.rs
@@ -71,7 +71,6 @@ pub fn generate_vello_scene(
         dom,
         scale,
         devtools: devtool_config,
-        scroll_offset: dom.scroll_offset,
     };
     generator.generate_vello_scene(scene);
 
@@ -90,7 +89,6 @@ pub struct VelloSceneGenerator<'dom> {
     dom: &'dom Document,
     scale: f64,
     devtools: Devtools,
-    scroll_offset: f64,
 }
 
 impl<'dom> VelloSceneGenerator<'dom> {
@@ -116,10 +114,7 @@ impl<'dom> VelloSceneGenerator<'dom> {
         self.render_element(
             scene,
             self.dom.as_ref().root_element().id,
-            Point {
-                x: 0.0,
-                y: self.scroll_offset,
-            },
+            Point { x: 0.0, y: 0.0 },
         );
 
         // Render debug overlay
@@ -163,8 +158,6 @@ impl<'dom> VelloSceneGenerator<'dom> {
             abs_x += x;
             abs_y += y;
         }
-
-        abs_y += self.scroll_offset as f32;
 
         // Hack: scale factor
         let abs_x = f64::from(abs_x) * scale;
@@ -314,8 +307,10 @@ impl<'dom> VelloSceneGenerator<'dom> {
 
         // TODO: account for overflow_x vs overflow_y
         let styles = &element.primary_styles().unwrap();
-        let overflow = styles.get_box().overflow_x;
-        let should_clip = !matches!(overflow, Overflow::Visible);
+        let overflow_x = styles.get_box().overflow_x;
+        let overflow_y = styles.get_box().overflow_y;
+        let should_clip =
+            !matches!(overflow_x, Overflow::Visible) || !matches!(overflow_y, Overflow::Visible);
         let clips_available = CLIPS_USED.load(atomic::Ordering::SeqCst) <= CLIP_LIMIT;
 
         // Apply padding/border offset to inline root
@@ -354,12 +349,26 @@ impl<'dom> VelloSceneGenerator<'dom> {
             CLIP_DEPTH_USED.fetch_max(depth, atomic::Ordering::SeqCst);
         }
 
-        let cx = self.element_cx(element, location);
+        let mut cx = self.element_cx(element, location);
         cx.stroke_effects(scene);
         cx.stroke_outline(scene);
         cx.stroke_frame(scene);
         cx.stroke_border(scene);
         cx.stroke_devtools(scene);
+
+        //Now that background has been drawn, offset pos and cx in order to draw our contents scrolled
+        let pos = Point {
+            x: pos.x - element.scroll_offset.x,
+            y: pos.y - element.scroll_offset.y,
+        };
+        cx.pos = Point {
+            x: cx.pos.x - element.scroll_offset.x,
+            y: cx.pos.y - element.scroll_offset.y,
+        };
+        cx.transform = cx.transform.then_translate(Vec2 {
+            x: cx.transform.translation().x - element.scroll_offset.x,
+            y: cx.transform.translation().y - element.scroll_offset.y,
+        });
         cx.draw_image(scene);
         cx.draw_svg(scene);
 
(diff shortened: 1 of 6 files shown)


jkelleyrtp:
Thanks for the PR! Nico's currently on break so I'll take a look at it, but it'll take a moment :)

cfraz89:
No problems

packages/blitz/src/renderer/render.rs
```diff
@@ -354,12 +349,26 @@ impl<'dom> VelloSceneGenerator<'dom> {
...
 
-        let cx = self.element_cx(element, location);
+        let mut cx = self.element_cx(element, location);
         cx.stroke_effects(scene);
         cx.stroke_outline(scene);
         cx.stroke_frame(scene);
         cx.stroke_border(scene);
         cx.stroke_devtools(scene);
+
+        //Now that background has been drawn, offset pos and cx in order to draw our contents scrolled
```
nicoburns, line 359: Nit: Comment formatting (please add space between `//` and `Now`)

packages/dom/src/node.rs
```diff
@@ -797,6 +800,50 @@ impl Node {
...
+        self.scroll_offset.y += y;
+
+        //If we're past our scroll bounds, transfer scrolling to parent
+        if self.scroll_offset.x < 0.0
+            || self.scroll_offset.x > self.final_layout.scroll_width() as f64
+            || self.scroll_offset.y < 0.0
+            || self.scroll_offset.y > self.final_layout.scroll_height() as f64
+        {
+            if let Some(parent) = self.parent {
+                unsafe {
```
nicoburns, line 817: If you moved the scroll_by method to the `Document` then you could avoid the `unsafe` here.

packages/dom/src/node.rs
```diff
@@ -797,6 +800,50 @@ impl Node {
...
+    /// Will bubble scrolling up to parent node once it can no longer scroll further
+    pub fn scroll_by(&mut self, x: f64, y: f64) {
+        self.scroll_offset.x += x;
+        self.scroll_offset.y += y;
+
+        //If we're past our scroll bounds, transfer scrolling to parent
+        if self.scroll_offset.x < 0.0
+            || self.scroll_offset.x > self.final_layout.scroll_width() as f64
+            || self.scroll_offset.y < 0.0
+            || self.scroll_offset.y > self.final_layout.scroll_height() as f64
```
nicoburns, line 814: This check should occur before updating the node's scroll offset. It should also allow *partial* transfer of the scroll delta to the parent. For example, if `scroll_offset.y` is 10 and x is `-20` then the current node should scroll by `-10` to scroll offset `0`, and the parent should be scrolled by `10`.

nicoburns: I believe if you did this then you wouldn't need to clamp.

packages/dom/src/document.rs
```diff
@@ -739,31 +744,6 @@ impl Document {
...
-    pub fn scroll_by(&mut self, px: f64) {
-        // Invert scrolling on macos
-        #[cfg(target_os = "macos")]
-        {
-            self.scroll_offset += px;
-        }
-        #[cfg(not(target_os = "macos"))]
-        {
-            self.scroll_offset -= px;
-        }
```
nicoburns, line 751: I think we still need this. This branch is scrolling in the incorrect direction for me. I am unsure if the inversion is macOS specific or scrolling needs to be inverted on all platforms.

packages/dom/src/document.rs
```diff
@@ -697,12 +692,22 @@ impl Document {
...
+
+        // Root element's size is viewport size
+        let root_element = self.get_node_mut(self.root_element().id).unwrap();
+
+        root_element.final_layout.size = taffy::Size {
+            width: size.width.to_f32_px(),
+            height: size.height.to_f32_px(),
+        };
+
+        root_element.clamp_scroll_offset();
```
nicoburns, line 710: I believe that the way the web works is that there is an implicit "viewport" node that lives above the HTML element that has a fixed size, with the HTML element itself being a normal node. This is probably close enough for the time being, although I don't like overriding the layout size (how does this work when generating "full height" screenshots?)

nicoburns: It would be good to have a way to leave the viewport size undefined (determined by layout) as this is used to render "full height screenshots" in the screenshot example (compare output of `cargo run --release --example screenshot https://servo.org` on this branch vs. main)

packages/dioxus-blitz/src/window.rs
```diff
@@ -409,6 +409,7 @@ impl<Doc: DocumentLike> View<Doc> {
             WindowEvent::Focused(_) => {}
 
             // Touch and motion events
+            //Todo implement touch scrolling
```
nicoburns, line 412: Nit: Comment formatting (please add space between `//` and `Todo`)

nicoburns (changes requested):
Thanks for the PR! Overall this looks like the correct approach, but I have few review comments

(comments prefixed with `Nit` are minor style issues that I would prefer to be addressed but will not block merging)

packages/dom/src/document.rs
cfraz89: Ok, I realised it was scrolling the wrong way on my gnome desktop too, so in the updated scrolling code I've inverted it. Could you double check it's scrolling the right way on mac now?

packages/dom/src/document.rs
cfraz89: Ok I've kind of restored the old behaviour where we keep a scroll property on the document, which gets bubbled up to from the root node. That seems to be the simplest way to avoid this issue for now, as I don't think we need to go so far as to actually have a viewport node in the tree yet.

cfraz89:
Thanks for the feedback, all makes sense. I revamped the PR:

- Scrolling behaviour is all implemented in document now
- Implemented the partial bubbling of scroll up to parent node, then to document scroll
- Removed locking root element size to viewport size, and instead track scroll on document again
- Flipped scroll direction 
- use kurbo::Point instead of taffy::Point to store scroll coords, realised that taffy::Point has a bit of a css slant
- Fixed comment formatting

nicoburns (changes requested):

packages/dioxus-blitz/src/window.rs
```diff
@@ -191,7 +191,7 @@ impl<Doc: DocumentLike> View<Doc> {
 
     pub fn mouse_move(&mut self, x: f32, y: f32) -> bool {
         let dom_x = x / self.viewport.zoom();
-        let dom_y = (y - self.dom.as_ref().scroll_offset as f32) / self.viewport.zoom();
+        let dom_y = y / self.viewport.zoom();
```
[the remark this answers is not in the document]
cfraz89: I see, have addressed this now, updated the scroll example to reflect it as well.

PR #122, opened 2024-08-20
Implement box-shadow drawing

cfraz89:
Adds support for rendering box-shadow, using the [blur-rect](https://github.com/msiglreith/vello/tree/blur-rect) branch of vello.

Also updates the tailwind example to render `#cool` with a black shadow, and #cool-inset with an inset shadow, to demo support.

*Caveats*
- Linked to a fork of vello
- Vello support is only of a single border radius, so it averages all corner's border radii to get the shadow radius
- Ignores the spread radius parameter, and only uses blur radius. Not sure how spread radius is meant to be handled?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,17 +12,17 @@ style_config = { git = "https://github.com/dioxuslabs/stylo", branch = "enable-t
 style_traits = { git = "https://github.com/dioxuslabs/stylo", branch = "enable-table-moz-center-style-adjust" }
 style_dom = { git = "https://github.com/dioxuslabs/stylo", package = "dom", branch = "enable-table-moz-center-style-adjust" }
 selectors = { git = "https://github.com/dioxuslabs/stylo", branch = "enable-table-moz-center-style-adjust" }
-html5ever = "0.27" # needs to match stylo markup5ever version
+html5ever = "0.27"                                                                                                            # needs to match stylo markup5ever version
 taffy = { git = "https://github.com/dioxuslabs/taffy", rev = "950a0eb1322f15e5d1083f4793b55d52061718de" }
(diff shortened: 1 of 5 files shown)


jkelleyrtp (approved):
Looks good! I think your formatter is doing some weird stuff in the Cargo.toml.

I'm going to wait a day or two to merge this in case the corresponding PR on vello gets merged - and then we can just run from their `main` branch instead of the actual blur branch itself.

Also, I'm not sure if they've implemented the varying-corner-radius logic yet. I believe not, and it'll require a new shader on their end to support it.

Whenever that lands, we need to update the radius here to match the actual multi-radius border logic that gets merged.

https://github.com/linebender/vello/pull/665

cfraz89:
Sounds good, I've reverted the formatting on the Cargo.toml.

nicoburns:
Does this fix hovering the search box in the "google" example? (seach box should have shadow on hover, currently it visually disappears instead).

The vello PR has now landed. We can merge this if:

- It is updated to use vello's `main` branch (some changes may be required)
- The main vello version in `Cargo.toml` is updated rather than a patch applied. This may require forking vello_svg to have a compatible version.

cfraz89:
> Does this fix hovering the search box in the "google" example? (seach box should have shadow on hover, currently it visually disappears instead).

It does :smiley: 

`
Screenshot from 2024-08-29 14-42-15
`

> The vello PR has now landed. We can merge this if:
> 
> * It is updated to use vello's `main` branch (some changes may be required)
> * The main vello version in `Cargo.toml` is updated rather than a patch applied. This may require forking vello_svg to have a compatible version.

Done. I've made a personal fork of vello_svg to depend on the updated Vello, not sure if thats acceptable or you'd rather update the fork at https://github.com/DioxusLabs/vello_svg

Issue #119, opened 2024-08-15
Roadmap


nicoburns:

savanesoff:
Hi 👋 

Really interesting project! 🚀 

Very silly question: Does the browser run JS?

This would be very useful for running web apps on an embedded devices, like STBs, Vehicle Infotainment systems etc... where browser bloat is a HUGE issue!

Many companies take off-the-shelf Browsers like `Chromium` (bloated) or `Cobalt` (out of date entirely, but trimmed down) to develop Web apps (react, svelt based) with a heavy use of JS.

If not, is it in the cards?

nicoburns:
> Very silly question: Does the browser run JS?
> 
> If not, is it in the cards?

@savanesoff It's not a silly question. The answer is that no it doesn't, but that it might be on the cards.

Architecturally we are setup for a JS engine (or engines!) to slot in (and we do already implement event handling and dynamic DOM updates) It's not a development priority for us in 2025 (we are prioritising the "core" renderer and the dioxus and rust APIs), but it could become one later (or sooner if there is an interested party that could bring development resources to work on the feature).

Aworldc:
> > Very silly question: Does the browser run JS?
> > If not, is it in the cards?
> 
> [@savanesoff](https://github.com/savanesoff) It's not a silly question. The answer is that no it doesn't, but that it might be on the cards.
> 
> Architecturally we are setup for a JS engine (or engines!) to slot in (and we do already implement event handling and dynamic DOM updates) It's not a development priority for us in 2025 (we are prioritising the "core" renderer and the dioxus and rust APIs), but it could become one later (or sooner if there is an interested party that could bring development resources to work on the feature).

I am looking at doing this with [boa](boajs.dev), but I'm very new to the project and am having a hard time locating the rust apis that I could build it on. Like I found `blitz_html::HtmlDocument.as_mut<BaseDocument>` and `blitz_dom::BaseDocument.query_selector` but that's it.

nicoburns:
> I am looking at doing this with [boa](boajs.dev), but I'm very new to the project and am having a hard time locating the rust apis that I could build it on. Like I found `blitz_html::HtmlDocument.as_mut<BaseDocument>` and `blitz_dom::BaseDocument.query_selector` but that's it.

You could take a look at the dioxus-native crate. That's the code we currently have that does the most around dynamic updates and interactivity. But yes, there's a lot that would need to be built out to make this work. Including, I believe, reference counting for DOM nodes.

[...]

PR #123, opened 2024-08-20
Adding net crate

kokoISnoTarget:
This adds a `net` crate with sync and async networking Providers as mentioned in #119 and the discord.
There is currently a bug that loaded CSS isn't resolved properly, that is why this is a draft. 

Please comment any thoughts regarding the implementation.

### DIFF
--- a/packages/dioxus-blitz/Cargo.toml
+++ b/packages/dioxus-blitz/Cargo.toml
@@ -26,7 +26,7 @@ style = { workspace = true }
 tracing = { workspace = true, optional = true }
 blitz = { path = "../blitz" }
 blitz-dom = { path = "../dom" }
-blitz-net = { path = "../net" }
+blitz-net = { path = "../net", features = ["non_blocking"] }
 html-escape = "0.2.13"
 url = { version = "2.5.0", features = ["serde"] }
 ureq = "2.9"
--- a/packages/dioxus-blitz/Cargo.toml
+++ b/packages/dioxus-blitz/Cargo.toml
@@ -26,6 +26,8 @@ style = { workspace = true }
 tracing = { workspace = true, optional = true }
 blitz = { path = "../blitz" }
 blitz-dom = { path = "../dom" }
+blitz-net = { path = "../net" }
+html-escape = "0.2.13"
 url = { version = "2.5.0", features = ["serde"] }
 ureq = "2.9"
 rustc-hash = "1.1.0"
--- a/packages/dom/src/htmlsink.rs
+++ b/packages/dom/src/htmlsink.rs
@@ -1,9 +1,9 @@
 use std::borrow::Cow;
-use std::collections::HashSet;
-use std::sync::Arc;
+use std::collections::{HashMap, HashSet};
+use std::sync::{Arc, Mutex};
 
 use crate::node::{Attribute, ElementNodeData, ImageData, Node, NodeData, NodeSpecificData};
-use crate::util::ImageOrSvg;
+use crate::util::{Resource, NetProvider};
 use crate::Document;
 use html5ever::local_name;
 use html5ever::{
@@ -21,7 +21,7 @@ fn html5ever_to_blitz_attr(attr: html5ever::Attribute) -> Attribute {
     }
 }
 
-pub struct DocumentHtmlParser<'a> {
+pub struct DocumentHtmlParser<'a, N> {
     doc: &'a mut Document,
 
     style_nodes: Vec<usize>,
@@ -31,20 +31,23 @@ pub struct DocumentHtmlParser<'a> {
 
     /// The document's quirks mode.
     pub quirks_mode: QuirksMode,
+    
+    net_provider: N,
 }
 
-impl<'a> DocumentHtmlParser<'a> {
-    pub fn new(doc: &mut Document) -> DocumentHtmlParser {
+impl<'a, N: NetProvider<usize, Resource>> DocumentHtmlParser<'a, N> {
+    pub fn new(doc: &mut Document, net_provider: N) -> DocumentHtmlParser<N> {
         DocumentHtmlParser {
             doc,
             style_nodes: Vec::new(),
             errors: Vec::new(),
             quirks_mode: QuirksMode::NoQuirks,
+            net_provider,
         }
     }
 
-    pub fn parse_into_doc<'d>(doc: &'d mut Document, html: &str) -> &'d mut Document {
-        let sink = Self::new(doc);
+    pub fn parse_into_doc<'d>(doc: &'d mut Document, net: N, html: &str) -> &'d mut Document {
+        let sink = Self::new(doc, net);
         html5ever::parse_document(sink, Default::default())
             .from_utf8()
             .read_from(&mut html.as_bytes())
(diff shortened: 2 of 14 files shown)


nicoburns:
Oh wow. This PR makes Blitz feel *fast* (or at least, it did before the workaround was added). Would love to get this merged soon.

The stylesheet bug is pretty bad though. We'll need to get to the bottom of that. It looks like it might only be inline nodes that styles aren't applying to, but I'm not quite sure why that would be as it looks like that's getting rebuilt every frame.

packages/dom/src/htmlsink.rs
```diff
@@ -31,20 +30,23 @@ pub struct DocumentHtmlParser<'a> {
 
     /// The document's quirks mode.
     pub quirks_mode: QuirksMode,
+
+    net_provider: N,
```
nicoburns, line 34: I think we should make this a `Box<dyn NetProvider<Resource>>`, then we could remove the generic parameter which would simplify the code and also ought to help compile times.

packages/net/src/provider/non_blocking.rs
```diff
@@ -0,0 +1,117 @@
...
+    ) {
+        self.rt.clone().spawn(async move {
+            let mut interval = tokio::time::interval(Duration::from_millis(100));
+
+            'thread: loop {
+                interval.tick().await;
+                while let Some(ir) = self.futures.lock().await.next().await {
+                    match ir {
+                        Ok(Ok(t)) => {
+                            let e = event_loop_proxy.send_event((window_id, t).into());
```
nicoburns, line 43: The `net` crate should not depend on `winit`.

You should use a channel here (probably a tokio [`mpsc::channel`](https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html)) in the API of the net crate, listen for messages from that channel in the top-level crate (currently the dioxus-blitz crate).

Alternatively you could have the net crate accept a closure to call on success and have that closure wrap the event loop proxy.

`WindowId` can be coerced into a u64.

packages/net/src/provider/non_blocking.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+impl<T: Send + Sync + 'static> AsyncProvider<T> {
+    pub fn new(rt: &Runtime) -> Self {
+        Self {
+            rt: rt.handle().clone(),
+            client: Client::new(),
+            futures: Mutex::new(FuturesUnordered::new()),
+        }
+    }
+    pub fn resolve<P: From<(WindowId, T)> + Send>(
```
nicoburns, line 30: If you have `fetch` accept the `WindowId` and the callback for the result directly, then I think this resolve function can be removed.

packages/net/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+use url::Url;
+
+pub use http::Method;
+
+pub use provider::*;
+
+#[cfg(any(feature = "blocking", feature = "non_blocking"))]
+const USER_AGENT: &str = "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/81.0";
+
+pub trait NetProvider<T> {
```
nicoburns, line 13: Can we move the `NetProvider` and `RequestHandler` traits into a new `blitz-traits` crate? That would allow `blitz-dom` to depend on the trait without depending on `blitz-net` (and vice versa). There are a couple of other types and traits that could also move into that crate (for similar reasons), but I think that would be best done as a separate PR.

I believe this would also allow us to avoid feature flags and have sync and async net providers in entirely separate crates.

packages/net/src/provider/blocking.rs
```diff
@@ -0,0 +1,74 @@
+use crate::{NetProvider, RequestHandler, USER_AGENT};
+use std::cell::RefCell;
+use std::io::Read;
+use thiserror::Error;
+use url::Url;
+
+const FILE_SIZE_LIMIT: u64 = 1_000_000_000; // 1GB
+
+pub struct SyncProvider<T>(pub RefCell<Vec<T>>);
```
nicoburns, line 9: Not sure what I think about the blocking client. I think a `ureq` option is a nice-to-have, but I feel like even simple use cases of Blitz will want to fetch network requests in parallel (which would imply using a ureq with a thread pool), and I'm not sure if it's worth the maintenance cost at this stage (we could also add a blocking wrapper around the non blocking client for cases like the screenshot example).

Is binary size optimisation the motivation for including a ureq-based client?

packages/net/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+{
+    fn fetch<H>(&self, url: Url, handler: H)
+    where
+        H: RequestHandler<T>,
+    {
+        self.deref().fetch(url, handler)
+    }
+}
+
+pub trait RequestHandler<T>: Send + Sync + 'static {
```
nicoburns, line 32: I feel like this abstraction could be simplified by having the Request type/trait contain only plain data. For example, it could contain the url/method and then a plain enum `enum Kind { Css, Image, Svg }`. The Request could then be passed back to the Document along with the response data, and the Document could have a `match` statement to parse the data. I feel like that would be simpler than having the net crate "know" about different resource types. Such a simple type could live in the "traits" crate so that "net" and "dom" don't have to depend on each other.

nicoburns (changes requested):
Looks like you've found the invalidation bug - nice!

I've put some review comments which I thought I had already sent, but looks like I didn't press the button. Please feel free to push back if you disagree with any of my comments: I think this is somewhere where there is quite a large design space and probably many good options.

One thing I am quite keen on is to minimise dependencies between crates. As such I would like the `net` crate to be as close to "url in, bytes out" as possible.

packages/net/src/provider/non_blocking.rs
kokoISnoTarget: The problem is that basically all fetching is done before the window (and the WindowId) exist.

packages/net/src/provider/non_blocking.rs
kokoISnoTarget: I'm not sure how I would use a channel with the winit runtime.

packages/net/src/lib.rs
kokoISnoTarget: The advantage of my current approach is that e.g. the CSS parsing can be done directly on the fetch thread. But changing to a "the document just gets the data" is pretty simple.

packages/net/src/provider/non_blocking.rs
nicoburns: I think this could be solved by giving the Document itself an ID The ID (perhaps just a simple sequential one?) could be generated either by the high-level glue code (or by `Document::new()`) and then the high-level code could keep track of which windows/documents that ID corresponds to.

This seems like a good idea anyway as in future we'll likely want to support things like replacing the Document associated with a Window (e.g. when clicking on a link), and having sub-Documents within a Document (e.g. for iframes, or using DioxusDocument wrapping an HtmlDocument to implement a browser ui).

packages/dioxus-blitz/Cargo.toml
```diff
@@ -26,6 +26,9 @@ style = { workspace = true }
 tracing = { workspace = true, optional = true }
 blitz = { path = "../blitz" }
 blitz-dom = { path = "../dom" }
+blitz-net = { path = "../net" }
+blitz-traits = { path = "../traits" }
+html-escape = "0.2.13"
```
nicoburns, line 31: Is `html-escape` actually used in this crate?

nicoburns (commented):
I still have quibbles about the structure of things here, but nothing major and I would now be keen to land this soon. The main blocker now being resolving the conflicts with the `main` branch.

nicoburns:
Regarding the `@font-face` support. It would also be cool to port the `woff` support from https://github.com/DioxusLabs/blitz/pull/107 as a lot of webfonts seem to be in woff format. But I wouldn't block this PR on that: that could be done in a followup.

packages/dioxus-blitz/Cargo.toml
kokoISnoTarget: Nope, forgot to remove it.

packages/net/src/lib.rs
nicoburns: I had in mind that the `net` crate would be a tokio executor in a single thread, and that further processing could be pushed onto a separate threadpool. But let's go with your approach for the time being.

nicoburns (approved):

Issue #94, opened 2024-07-09
Checkbox element


matthunz:
Should we support `input { type: "checkbox" }` by displaying a default input widget?

I think the big question here is what should the default styling look like :eyes: I'm thinking now we might want to match browser implementations somewhat close, but we may have some creative freedom here.

One idea I had is re-using  the `masonry` [Checkbox](https://docs.rs/masonry/latest/masonry/widget/struct.Checkbox.html) widget here

nicoburns:
Read-only checkboxes would also be useful for rendering checkbox lists in Github READMEs. Our markdown library already supports generating them. Perhaps the most expedient path forwards would be to implement this custom but use the Masonry Checkbox's paint implementation (which is of course vello-based anyway).

nicoburns:
Closed by #125

PR #125, opened 2024-09-01
Checkbox inputs

cfraz89:
Adds support for input type="checkbox", as requested in #94 . Makes checkboxes send `oninput` event when clicked. If event isn't bound (there's no data-dioxus-id), they will behave as an unmanaged input, and handle checked state internally. If it is, they will send an oninput event with a FormData argument.

Adds a 'form' example to demonstrate.

Caveats:
- They should be drawing their colors using the css property 'accent-color', but I couldnt figure out how to look that up, so they're using just 'color' for now.
- There's something up with the vertical alignment when rendering labels with the text 'inline'. Check the second checkbox in the example, or try removing the 'inline-block' from label styling in the css. In chrome and ff, the labels render the same vertically whether inline or inline-block.
- The FormEvent has `valid: false`, as its calculated as 'true' if value is empty? However I've set value to the value of the toggled element. Not sure if this is correct
- Checking a checkbox send 'onclick' and 'oninput' events. Not sure if it should send 'onchanged' too? Didnt see any mention of that event in dioxus docs.

image

### DIFF
--- a/packages/blitz/src/renderer/render.rs
+++ b/packages/blitz/src/renderer/render.rs
@@ -40,6 +40,7 @@ use style::{
 use image::{imageops::FilterType, DynamicImage};
 use parley::layout::PositionedLayoutItem;
 use taffy::prelude::Layout;
+use vello::kurbo::{BezPath, Cap, Join};
 use vello::{
     kurbo::{Affine, Point, Rect, Shape, Stroke, Vec2},
     peniko::{self, Brush, Color, Fill, Mix},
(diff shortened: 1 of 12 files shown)


packages/dom/src/layout/mod.rs
```diff
@@ -144,10 +144,36 @@ impl LayoutPartialTree for Document {
...
+                            // if the input type is hidden, hide it
+                            Some("hidden") => {
+                                node.style.display = Display::None;
+                                return taffy::LayoutOutput::HIDDEN;
+                            }
+                            Some("checkbox") => {
+                                return compute_leaf_layout(
+                                    inputs,
+                                    &node.style,
+                                    |_known_size, _available_space| {
```
nicoburns, line 157: I suspect that the vertical alignment issue you are seeing is due to taffy/parley assuming the bottom of the node's border-box as it's baseline. Whereas for checkboxes it probably ought to ignore the margin and use the bottom of the actual checkbox as the baseline.

We will need to add support for inline-box baselines to parley to make that work properly, but a good first step would be to modify the `first_baselines.y` field of the `taffy::LayoutOutput` returned by `compute_leaf_layout` to set the baseline to the bottom of the checkbox (height of checkbox + top padding/border/margin).

packages/dom/src/document.rs
```diff
@@ -243,6 +265,71 @@ impl Document {
...
+                        return None;
+                    }
+                    let id = element_data.id.as_ref()?;
+                    if *id == *target_element_dom_id {
+                        Some(node)
+                    } else {
+                        None
+                    }
+                })
+                .collect()
```
nicoburns, line 292: We probably ought to maintain a global ID map so this can be looked up much more quickly? I wouldn't block this PR on that, but perhaps an "iterate over nodes with a given id" method could be added to `Document`?

packages/blitz/src/renderer/render.rs
```diff
@@ -1091,4 +1093,61 @@ impl ElementCx<'_> {
...
+        if self.element.local_name() == "input"
+            && matches!(self.element.attr(local_name!("type")), Some("checkbox"))
+        {
+            let checked: bool = self
+                .element
+                .attr(local_name!("checked"))
+                .and_then(|c| c.parse().ok())
+                .unwrap_or_default();
+
+            // TODO this should be coming from css accent-color, but I couldn't find how to retrieve it
```
nicoburns, line 1107: `accent-color` can be enabled here: https://github.com/servo/stylo/blob/main/style/properties/longhands/inherited_ui.mako.rs#L94 by changing `engines="gecko"` to `engines="gecko servo"`.

Such a change could be submitted to the DioxusLabs fork of Stylo. We are currently using the [`enable-table-moz-center-style-adjust`](https://github.com/DioxusLabs/stylo/tree/enable-table-moz-center-style-adjust) branch. So changes should be based on that branch (we should probably be more principled about using a `blitz` branch or similar).

packages/blitz/src/renderer/render.rs
```diff
@@ -1091,4 +1093,61 @@ impl ElementCx<'_> {
...
+
+    fn draw_input(&self, scene: &mut Scene) {
+        if self.element.local_name() == "input"
+            && matches!(self.element.attr(local_name!("type")), Some("checkbox"))
+        {
+            let checked: bool = self
+                .element
+                .attr(local_name!("checked"))
+                .and_then(|c| c.parse().ok())
+                .unwrap_or_default();
```
nicoburns, line 1105: `checked` is a boolean attribute, so this should just check for the presence of the `checked` attribute in the array. If it is present then `checked` is `true` else `false` (even the empty string or "false" counts as `true`).

See: https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

packages/dom/src/document.rs
```diff
@@ -243,6 +265,71 @@ impl Document {
...
+            };
+            el.attrs.push(attr);
+            el.attrs
+                .iter_mut()
+                .find(|attr| attr.name.local == local_name!("checked"))
+                .unwrap()
+        };
+        let checked = checked_attr.value.parse().unwrap_or(false);
+        checked_attr.value = (!checked).to_string();
+    }
```
nicoburns, line 331: This needs to actually remove the attribute entirely when unchecking and add it when checking. I suggest using the  empty string for the value to avoid an allocation.

packages/dom/src/document.rs
```diff
@@ -124,21 +125,42 @@ impl DocumentLike for Document {
...
 
                         self.set_focus_to(hit.node_id);
+                    } else if el.name.local == local_name!("input")
+                        && matches!(el.attr(local_name!("type")), Some("checkbox"))
+                    {
+                        Document::toggle_checkbox(el);
+                        self.set_focus_to(hit.node_id);
+                    }
+                    // Clicking labels triggers click, and possibly input event, of associated input
+                    else if el.name.local == local_name!("label") {
```
nicoburns, line 151: Nice that you've implemented label!

nicoburns (changes requested):
> They should be drawing their colors using the css property 'accent-color', but I couldnt figure out how to look that up, so they're using just 'color' for now.

See comments on relevant code.

> There's something up with the vertical alignment when rendering labels with the text 'inline'. Check the second checkbox in the example, or try removing the 'inline-block' from label styling in the css. In chrome and ff, the labels render the same vertically whether inline or inline-block.

Weird that the two checkboxes are rendering differently, but let's not worry too much about this for now.

> The FormEvent has valid: false, as its calculated as 'true' if value is empty? However I've set value to the value of the toggled element. Not sure if this is correct

I believe it should be the literal `on` if the value is empty. Saw that in an MDN doc somewhere.

> Checking a checkbox send 'onclick' and 'oninput' events. Not sure if it should send 'onchanged' too? Didnt see any mention of that event in dioxus docs.

I believe browsers send `onchanged` for checkboxes, so we probably should too.

---

Some nice to haves (non-blocking):

- Implementing the `disabled` attribute, which would prevent clicks from taking effect and would also make the checkbox render with a greyed out look.
- Making the space bar toggle the checkbox if it is currently focussed.

nicoburns:
This PR seems to be breaking button layout on the `google` example.

`main` branch:

Screenshot 2024-09-05 at 15 03 03


This PR:

Screenshot 2024-09-05 at 15 03 31

nicoburns (approved):

PR #130, opened 2024-09-09
Port full gradient support to blitz

sepcnt:
This PR revives the long-stalled PR from https://github.com/jkelleyrtp/stylo-dioxus/pull/4.

It adds full gradient support. Note that due to Vello's behavior, it follows the SVG schema rather than the web schema by default ("transparent" is #00000000 rather than #FFFFFF00 in color mixing), so some gradients may differ.

### DIFF
--- a/examples/gradient.rs
+++ b/examples/gradient.rs
@@ -8,32 +8,83 @@ fn app() -> Element {
     rsx! {
         style { {CSS} }
         div {
-            // https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient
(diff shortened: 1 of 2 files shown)


nicoburns (approved):
Ah amazing. I was hoping that you would revive that PR! I am not able to verify the math, but this seems to fix the rendering of the linear gradient on servo.org (it now matches chrome), and even an incorrect rendering of the radial and conic gradients will be much better than what we currently have (which panics!), so I'm happy to trust that it is correct :)

PR #152, opened 2024-10-27
Refactor and document Viewport

matthunz:
* Replaces `Viewport::zoom` and `Viewport::set_zoom` with an exported field
* Removes `Viewport::scale_f64` convenience method
  - My thought here is to keep the codebase minimal while we're still building
* Docs and general refactors

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -151,8 +151,10 @@ impl DocumentLike for Document {
                     if let NodeSpecificData::TextInput(ref mut text_input_data) =
                         el.node_specific_data
                     {
-                        let x = (hit.x - content_box_offset.x) as f64 * self.viewport.scale_f64();
-                        let y = (hit.y - content_box_offset.y) as f64 * self.viewport.scale_f64();
+                        let x =
+                            (hit.x - content_box_offset.x) as f64 * self.viewport.scale() as f64;
+                        let y =
+                            (hit.y - content_box_offset.y) as f64 * self.viewport.scale() as f64;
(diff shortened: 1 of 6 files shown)


packages/blitz-dom/src/document.rs
```diff
@@ -1132,8 +1134,8 @@ impl Document {
     pub fn scroll_viewport_by(&mut self, x: f64, y: f64) {
         let content_size = self.root_element().final_layout.size;
         let new_scroll = (self.viewport_scroll.x - x, self.viewport_scroll.y - y);
-        let window_width = self.viewport.window_size.0 as f64 / self.viewport.scale() as f64;
-        let window_height = self.viewport.window_size.1 as f64 / self.viewport.scale() as f64;
+        let window_width = self.viewport.window_size.0 as f64 / self.viewport.hidpi_scale as f64;
+        let window_height = self.viewport.window_size.1 as f64 / self.viewport.hidpi_scale as f64;
```
nicoburns, line 1138: Was this change intentional? It's not equivalent to the previous code (which included the `zoom`).

If not, then IMO this is actually quite a good argument for having the `hidpi_scale` and` zoom` field private as they were before, as it prevents this kind of accidental usage of the wrong value.

If it is, then can you provide a justification for the change?

nicoburns (changes requested):
The documentation additions here are good. The impact of the other changes is IMO negligible (a few single-line methods on Viewport aren't exactly a maintenance burden), but I will accept them (with the requested changes) if you're keen for them.

nicoburns:
Closing as this now has a large number of conflicts.

PR #159, opened 2024-11-20
chore(refactor): improve code readability in renderer

kokoISnoTarget:
- Renames `element`  to `node`
- Renames `pos` to `box_position` and `content_position`
- Renames `size` to `content_size`
- Adds an `element` fields to `ElementCx` which is of type `ElementNodeData` because we `node.element_data().unwrap()` 6 times
- and move some code into their own functions  (`draw_text_input_text`, `draw_marker` and `draw_children`)


nicoburns (approved):
All looks good to me - thanks!

PR #166, opened 2024-11-29
Add prerequisites for (arch)linux, especially if running wayland

nerdachse:


nicoburns:
I've disabled the `libxdo` feature in `muda` by default. So in theory this should no longer be required. Could you check with latest `main`?

nerdachse:
I'll check with latest main asap by uninstalling xdo-tools from my machine again.

nerdachse:
Seems to work fine!

PR #167, opened 2024-11-30
Remove event loop spamming when new event arrives

DavidHusicka:
`new_event` method is called when a new event arrives on an event loop. Putting new event (in this case `BlitzEvent::Poll`) on the same event loop from this method causes this method to be called again. This ends up spamming event loop, 100% CPU utilization on some operating systems, and makes the `poll` "spin".


nicoburns (approved):
This makes sense. Which platform(s) are you seeing the high CPU usage on?

DavidHusicka:
> This makes sense. Which platform(s) are you seeing the high CPU usage on?

I have 100% CPU usage on a single core on Linux wayland. I already brought that up on Discord before.

https://discord.com/channels/899851952891002890/954257659597553664/1272540988057063525
https://discord.com/channels/899851952891002890/954257659597553664/1272824155674251357

PR #172, opened 2024-12-26
feat: Add mousedown and mouseup event

luoxiaozero:
Click events are triggered after the mouse is pressed and released, and on the same node.

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -172,6 +172,7 @@ impl DocumentLike for Document {
         let target_node_id = event.target;
 
         match event.data {
+            EventData::MouseDown { .. } | EventData::MouseUp { .. } => {}
             EventData::Click { x, y, .. } => {
                 let hit = self.hit(x, y);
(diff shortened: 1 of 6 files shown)


nerdachse:
Is unactive an English word? I am a non native speaker, sorry for the stupid question. I know inactive and to deactivate but unactive is new to me

luoxiaozero:
No, I just kept the same naming convention as [unhover](https://github.com/DioxusLabs/blitz/blob/6d3733d21555a6facc57009b5f85f7a0b8f69594/packages/blitz-dom/src/node.rs#L209). But I am very open to renaming variables.

nicoburns (approved):
We'll need to think much more carefully about the interaction between:

- Handling events in dioxus vs. in DOM
- Bubbling events
- Triggering events based on other events (e.g. mousedown+mouseup triggering click, click sometimes triggering input or focus)

But this seems ok for now.

nicoburns:
Be cool to get mousemove working as well. Then we could do mouse selection in text inputs!

PR #173, opened 2024-12-26
Add the ability to follow links

nerdachse:
For this we added another Variant to the BlitzEvent Enum and also made
 DocumentLike return an Option<DocumentEvent> which currently just has
 one Variant for following links.

 I consider this a working POC, which needs to be fleshed out.
 I didn't invest any time to make sure the DioxusSide still works. It
 most likely won't because of the change in the trait definition.

@nicoburns If this is about what you want, I am happy to work on this, as my time allows. If I've diverged too far from your vision, that's also fine!

It's a working POC. When I open the readme app and click a link, the Document is replaced.

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -229,8 +229,14 @@ impl DocumentLike for Document {
                         }
                     } else if el.name.local == local_name!("a") {
                         if let Some(href) = el.attr(local_name!("href")) {
-                            let url = resolve_url(&self.base_url, href);
-                            self.navigation_provider.navigate_new_page(url.into());
+                            if let Some(url) = resolve_url(&self.base_url, href) {
+                                self.navigation_provider.navigate_new_page(url.into());
+                            } else {
+                                println!(
+                                    "{href} is not parseable as a url. : {base_url:?}",
+                                    base_url = self.base_url
+                                )
+                            }
                         } else {
                             println!("Clicked link without href: {:?}", el.attrs());
                         }
@@ -604,7 +610,12 @@ impl Document {
     }
 
     pub fn resolve_url(&self, raw: &str) -> url::Url {
-        resolve_url(&self.base_url, raw)
+        resolve_url(&self.base_url, raw).unwrap_or_else(|| {
+            panic!(
+                "to be able to resolve {raw} with the base_url: {base_url:?}",
+                base_url = self.base_url
+            )
+        })
     }
 
     pub fn print_tree(&self) {
@@ -1240,11 +1251,10 @@ impl AsMut<Document> for Document {
     }
 }
 
-fn resolve_url(base_url: &Option<url::Url>, raw: &str) -> url::Url {
-    // TODO: this will bite use eventually. You can be sure on the www there is a myriad of links that
-    // cannot be parsed.
+fn resolve_url(base_url: &Option<url::Url>, raw: &str) -> Option<url::Url> {
     match base_url {
-        Some(base_url) => base_url.join(raw).unwrap(),
-        None => url::Url::parse(raw).unwrap(),
+        Some(base_url) => base_url.join(raw),
+        None => url::Url::parse(raw),
     }
+    .ok()
 }
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -70,14 +70,19 @@ impl FontMetricsProvider for DummyFontMetricsProvider {
     }
 }
 
+pub enum DocumentEvent {
+    ClickedLink(String),
+}
+
 pub trait DocumentLike: AsRef<Document> + AsMut<Document> + Into<Document> + 'static {
     fn poll(&mut self, _cx: std::task::Context) -> bool {
         // Default implementation does nothing
         false
     }
 
-    fn handle_event(&mut self, _event: RendererEvent) {
+    fn handle_event(&mut self, _event: RendererEvent) -> Option<DocumentEvent> {
         // Default implementation does nothing
+        None
     }
 
     fn as_any_mut(&mut self) -> &mut dyn Any {
@@ -168,7 +173,7 @@ fn make_device(viewport: &Viewport) -> Device {
 }
 
 impl DocumentLike for Document {
-    fn handle_event(&mut self, event: RendererEvent) {
+    fn handle_event(&mut self, event: RendererEvent) -> Option<DocumentEvent> {
         let target_node_id = event.target;
 
         match event.data {
@@ -182,13 +187,11 @@ impl DocumentLike for Document {
                         x: node.final_layout.padding.left + node.final_layout.border.left,
                         y: node.final_layout.padding.top + node.final_layout.border.top,
                     };
-                    let Some(el) = node.raw_dom_data.downcast_element_mut() else {
-                        return;
-                    };
+                    let el = node.raw_dom_data.downcast_element_mut()?;
 
                     let disabled = el.attr(local_name!("disabled")).is_some();
                     if disabled {
-                        return;
+                        return None;
                     }
 
                     if let NodeSpecificData::TextInput(ref mut text_input_data) =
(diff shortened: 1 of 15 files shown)


nicoburns:
@nerdachse I reckon we ought to create a `NavigationProvider` trait (not entirely set on the name) similar to the `NetProvider` trait. And then have the `Document` accept and store an `Arc<dyn NavigationProvider>`. That trait can have a method on it to navigate to the new page (and we could later extend it with functionality like "refresh page", "exit application", and maybe even "open new window"), and the implementation of that method can use the `proxy` as this PR already does.

I don't like the way the current implementation uses the return value, because I think it's non-obvious that this is what the return value would, and also because I think it is likely we will want to add more events in future (possibly even with a single handler generating more than one), and I think it would be difficult to allow for multiple events if using the return value.

nerdachse:
I agree, that's why I proposed two different implementation strategies.

I like that your solution has a specific Provider and not just the whole EventLoop exposed to the document.

I'll get to it as soon as I can!

nerdachse:
Still not finished for `Dioxus` side, but I want to iterate on this, to find a satisfying solution.

Happy to hear your feedback @nicoburns!

packages/blitz-shell/src/window.rs
```diff
@@ -270,7 +270,7 @@ impl<Doc: DocumentLike, Rend: DocumentRenderer> View<Doc, Rend> {
                         y: self.dom_mouse_pos.1,
                         mods: self.keyboard_modifiers,
                     },
-                });
+                })
```
nerdachse, line 273: Sorry for that one, I can fix it of course...

packages/blitz-dom/src/document.rs
```diff
@@ -222,6 +227,12 @@ impl DocumentLike for Document {
                             }
                             self.set_focus_to(node_id);
                         }
+                    } else if el.name.local == local_name!("a") {
+                        if let Some(href) = el.attr(local_name!("href")) {
+                            self.navigation_provider.navigate_new_page(href.to_owned());
```
nicoburns, line 232: We need to call `Document::resolve_url` (or equivalent functionality) on the href before sending it to the navigation provider in order to support relative links. I ran into borrow checker issues doing this. One solution might be to clone the base_url earlier in this method and move the resolution logic to a free functions (or wrap the base_url in a newtype with the method on it).

packages/blitz-dom/src/document.rs
```diff
@@ -222,6 +227,12 @@ impl DocumentLike for Document {
                             }
                             self.set_focus_to(node_id);
                         }
+                    } else if el.name.local == local_name!("a") {
+                        if let Some(href) = el.attr(local_name!("href")) {
```
nicoburns, line 231: So this is working. But it only works if you click the actual `<a>` element. It doesn't work if you click a child within the `<a>` (even a text node). We should somehow implement event bubbling to make this work.

nicoburns (changes requested):
The basic structure of this looks good. I have suggested a few tweaks in individual comments. Additionally compiling for some apps (dioxus-native, wpt runner) is broken which will need to be fixed before we can land this. You can check all apps by running `cargo build --workspace` (or checking the CI output).

packages/blitz-dom/src/document.rs
nerdachse: Actually we have a bigger, somehow related problem here.

Depending on the page, when I click an icon next to a link or so, the assertion that the node we clicked on and the node that fired the event are the same doesn't hold anymore and blitz panics.

Locally I've just removed that assertion, but maybe we should check if they are descendants of one another or so? Not sure.

Also on phone right now, so I don't know where exactly it originates.

packages/blitz-traits/src/navigation.rs
```diff
@@ -0,0 +1,16 @@
+use std::sync::Arc;
+
+pub type SharedNavigationProvider = Arc<dyn NavigationProvider>;
```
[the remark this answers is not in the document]
nerdachse: Yes, please 🥺

nerdachse:
> The basic structure of this looks good. I have suggested a few tweaks in individual comments. Additionally compiling for some apps (dioxus-native, wpt runner) is broken which will need to be fixed before we can land this. You can check all apps by running `cargo build --workspace` (or checking the CI output).

Sure, I just didn't want to waste time. Now that the path forward is clear, I'll fix it :)

packages/blitz-dom/src/document.rs
nerdachse: > So this is working. But it only works if you click the actual `<a>` element. It doesn't work if you click a child within the `<a>` (even a text node). We should somehow implement event bubbling to make this work.

```
thread 'main' panicked at packages/blitz-dom/src/document.rs:183:21:
assertion failed: hit.node_id == event.target
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I guess that's what you meant. I shouldn't write comments while doing groceries :D

packages/blitz-dom/src/document.rs
```diff
@@ -1224,3 +1239,12 @@ impl AsMut<Document> for Document {
         self
     }
 }
+
+fn resolve_url(base_url: &Option<url::Url>, raw: &str) -> url::Url {
+    // TODO: this will bite use eventually. You can be sure on the www there is a myriad of links that
```
nerdachse, line 1244: Note this. Not sure about the policy of adding TODOs.

packages/blitz/Cargo.toml
```diff
@@ -18,6 +18,7 @@ blitz-renderer-vello = { path = "../blitz-renderer-vello" }
 blitz-html = { path = "../blitz-html" }
 blitz-shell = { path = "../blitz-shell" }
 blitz-net = { path = "../blitz-net", optional = true }
+blitz-traits = { path = "../blitz-traits" }
```
nerdachse, line 21: @nicoburns not sure this is okay for you. But I am wondering how the net feature cfg could even compile, as it does an import of blitz-traits...

I mean this:

`#[cfg(not(feature = "net"))]
    let net_provider = {
        use blitz_traits::net::DummyNetProvider;
        Arc::new(DummyNetProvider::default())
    };`

packages/blitz-shell/src/event.rs
```diff
@@ -26,6 +26,9 @@ pub enum BlitzEvent {
 
     /// An arbitary event from the Blitz embedder
     Embedder(Arc<dyn Any + Send + Sync>),
+
+    /// Navigate to another URL (triggered by e.g. clicking a link)
+    Navigate(String),
```
nerdachse, line 31: Or we make this an URL here? I wasn't sure... can one navigate to something that's not an URL? The HTML spec is convoluted, I don't know even a 100th of the special cases...

packages/blitz-traits/src/navigation.rs
```diff
@@ -0,0 +1,12 @@
+/// A provider to enable a document to bubble up navigation events (e.g. clicking a link)
+pub trait NavigationProvider: Send + Sync + 'static {
```
nerdachse, line 2: It's now Send + Sync + 'static

packages/blitz-dom/src/document.rs
nicoburns: This seems fine, but there's a typo `use` instead of `us`. I guess probably we ought to make this return `Option<url>` and then callers can choose what to do if the URL doesn't parse (e.g. print a warning instead of panic, or pass on the unparsed string instead).

packages/blitz/Cargo.toml
nicoburns: I think this probably worked because the other `blitz-*` crates depend on `blitz-traits`. But there definitely should be an explicit dependency here.

packages/blitz-shell/src/event.rs
nicoburns: > can one navigate to something that's not an URL

Probably not on the web. But I'd be inclined to leave this as a `String` for now. Seems like it could be useful for people who want to make their own custom URL representations.

packages/blitz-dom/src/document.rs
nicoburns: I'm definitely hitting a case where nothing happens rather than a panic (open `https://servo.org` and try clicking the text in the header links). How are you triggering the panic?

nerdachse:
What this explicitly does not do is set the title of the window accordingly. So when you follow a link to another page, the title stays the current one.

nerdachse:
@nicoburns when you're ready to merge, please squash. Thanks :)

nicoburns (approved):

PR #176, opened 2025-01-07
Build requests via NetHandler

kokoISnoTarget:
This allows the `NetHandler` implementation to modify the request used to fetch.

This is required when we intend to implement forms in the future.

### DIFF
--- a/apps/wpt/src/net_provider.rs
+++ b/apps/wpt/src/net_provider.rs
@@ -1,4 +1,4 @@
-use blitz_traits::net::{BoxedHandler, Bytes, NetCallback, NetProvider, Url};
+use blitz_traits::net::{BoxedHandler, Bytes, NetCallback, NetProvider, Request};
 use data_url::DataUrl;
 use std::{
     path::{Path, PathBuf},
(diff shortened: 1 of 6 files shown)


nicoburns (commented):
Rather than having:
- `fetch` take a `Url`
- the `net` crate create a `Builder`
- and then giving the handler the opportunity to modify it

I think it would make more sense to just have `fetch` take a `Request` instead of the `Url`? What do you think?

kokoISnoTarget:
That's probably better in the long run, but the current net abstraction is flawed anyway (and needs a rewrite) as stated by you. And I think it's fine until then.

But I can change this if you want.

nicoburns (approved):
I'm not entirely sure about using our own Request type rather than just using the one from `http`. But perhaps it makes sense seeing as we support more than just `http`.

kokoISnoTarget:
> But perhaps it makes sense seeing as we support more than just http.

The Uri type used by http's Request doesn't fit well.

nicoburns:
> > But perhaps it makes sense seeing as we support more than just http.
> 
> The Uri type used by http's Request doesn't fit well.

Yeah, I agree to be honest.

PR #177, opened 2025-01-09
Give NavigationProvider options 

kokoISnoTarget:
This adds a type which represents a Navigation  
(https://html.spec.whatwg.org/multipage/browsing-the-web.html#beginning-navigation)
The `NavigationOptions` are expanded later when adding something like forms.

### DIFF
--- a/apps/readme/src/main.rs
+++ b/apps/readme/src/main.rs
@@ -3,7 +3,7 @@ mod readme_application;
 
 use blitz_html::HtmlDocument;
 use blitz_net::Provider;
-use blitz_traits::navigation::NavigationProvider;
+use blitz_traits::navigation::{NavigationOptions, NavigationProvider};
 use markdown::{markdown_to_html, BLITZ_MD_STYLES, GITHUB_MD_STYLES};
 use notify::{Error as NotifyError, Event as NotifyEvent, RecursiveMode, Watcher as _};
 use readme_application::{ReadmeApplication, ReadmeEvent};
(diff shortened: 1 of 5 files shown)


nicoburns:
> The NavigationOptions are expanded later when adding something like forms.

I feel like I would rather add them when adding forms so that we can prove the implementation with a real use case. If you're interested in implementing form submission then maybe you should just implement form submission? I don't think there is anything blocking that.

I would imagine implementation would look like:
- `Navigate` event takes a `Request` rather than a `Url`
- Form submission is triggered on:
   - Click of `<input type="submit" />`
   - Enter key press when single-line text-field is focussed
- Write a submission function that:
   - Iterates up the tree from the element that triggered the event looking for a `<form>` element
   - (if no `<form>` element is found then do nothing)
   - Reads the `action` and `method` from the `<form>`
   - Searches the subtree under the `<form>` for input elements (we currently support: text, textarea, radio, checkbox, and hidden).
   - Composes a `url` and `body` from form attrs and input values.
   - Generates a `Navigate` event.
- Initial implementation:
   - Wouldn't necessarily need to support all types (maybe just text and hidden?)
   - Wouldn't necessarily need to generate a "submit" event for Dioxus Native
   - Could assume `enctype` of `application/x-www-form-urlencoded`. But should maybe check for `multipart/form-data` and do nothing / log warning if it is set.

Although of course these features could be implemented (although probably no point doing multipart encoding until we have file inputs)

kokoISnoTarget:
Yea, I'm doing that currently (trying to follow the spec)

PR #180, opened 2025-01-09
Hover Implementation

wheregmis:
- [ ] Might need to clean up the example
- [ ] I have few comments where i might need some inputs

### DIFF
--- a/packages/blitz-dom/src/events/mod.rs
+++ b/packages/blitz-dom/src/events/mod.rs
@@ -1,8 +1,10 @@
+mod hover;
 mod ime;
 mod keyboard;
 mod mouse;
 
 use blitz_traits::{DomEvent, DomEventData};
+pub(crate) use hover::handle_hover;
 pub(crate) use ime::handle_ime_event;
 pub(crate) use keyboard::handle_keypress;
 pub(crate) use mouse::handle_click;
@@ -15,7 +17,9 @@ pub(crate) fn handle_event(doc: &mut BaseDocument, event: DomEvent) {
     match event.data {
         DomEventData::MouseDown(_) => {}
         DomEventData::MouseUp(_) => {}
-        DomEventData::Hover => {}
+        DomEventData::Hover(event) => {
+            handle_hover(doc, target_node_id, event.x, event.y);
+        }
         DomEventData::Click(event) => {
             handle_click(doc, target_node_id, event.x, event.y);
         }
--- a/packages/blitz-shell/src/window.rs
+++ b/packages/blitz-shell/src/window.rs
@@ -3,7 +3,7 @@ use crate::convert_events::{
 };
 use crate::event::{create_waker, BlitzShellEvent};
 use blitz_dom::BaseDocument;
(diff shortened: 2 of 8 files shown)


packages/blitz-shell/src/window.rs
```diff
@@ -249,12 +249,36 @@ impl<Doc: Document<Doc = D>, Rend: DocumentRenderer<Doc = D>> View<Doc, Rend> {
...
         self.dom_mouse_pos = (dom_x, dom_y);
-        self.doc.as_mut().set_hover_to(dom_x, dom_y)
+
+        // Get previous and new hover nodes
+        let previous_hover_node = self.doc.as_ref().get_hover_node_id();
+        let hover_changed = self.doc.as_mut().set_hover_to(dom_x, dom_y);
+        let current_hover_node = self.doc.as_ref().get_hover_node_id();
+
+        if previous_hover_node != current_hover_node {
+            // ! No Idea how Handle mouseleave for previous node
```
wheregmis, line 264: I have no idea if we need to handle this or not. Any direction or help is appreciated

packages/blitz-dom/src/events/mod.rs
```diff
@@ -15,7 +17,9 @@ pub(crate) fn handle_event(doc: &mut BaseDocument, event: DomEvent) {
     match event.data {
         DomEventData::MouseDown(_) => {}
         DomEventData::MouseUp(_) => {}
-        DomEventData::Hover => {}
+        DomEventData::Hover(event) => {
+            handle_hover(doc, target_node_id, event.x, event.y);
+        }
```
nicoburns, line 22: Can we rename this `MouseOver`?

packages/blitz-shell/src/window.rs
nicoburns: We definitely want to handle this. We'll want to create a new variant of the `DomEvent` struct to dispatch.

nicoburns (commented):
This looks like a decent change. However, I'm I'm going to try to land https://github.com/DioxusLabs/blitz/pull/179 first as I think it's better to make these architectural changes before adding more events.

I think we'll want to move to a model where `blitz-shell` just passes on the `mousemove` event, and detection of `mouseover`/`mouseenter`/`mouseout`/`mouseleave` happens in `blitz-dom`.

wheregmis:
Sure, i will wait for #179 to land. And can try to rebase to it. Till that time i will try to poke the codes during my spare time to know abit more about the codebase. Appreciate the direction and review

wheregmis:
Hi @nicoburns , I did an attempt to make mouseover/mouseenter/mouseout/mouseleave happening in blitz-dom but i am not able to make it happen. We might need to handle the events on Document of blitzshell for the event, which i dont have access to on the blitz-dom. Not sure if i am able to explain it properly. Any guidance will be appreciated.

wheregmis:
> Is this somehow related to missing an "active" state on buttons?
> 
> I started a discussion here: [DioxusLabs/dioxus#3622](https://github.com/DioxusLabs/dioxus/discussions/3622)

Its not, blitz is currently being developed and is not the default renderer for dioxus.

PR #182, opened 2025-01-20
feat: Support text box mouse selection

luoxiaozero:
This is a smaller PR than #179, which only supports selecting the input box text with the mouse.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,6 +84,7 @@ futures-intrusive = "0.5.0"
 thiserror = "1.0.63"
 pollster = "0.4"
 smol_str = "0.2"
+bitflags = "2.8.0"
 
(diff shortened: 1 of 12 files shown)


nicoburns (approved):
This looks good to me. Event system in general is still quite messy, but that was true before this PR, and this isn't making anything any worse.

The performance of mouse selection is noticeably worse in the `todomvc` example than it is opening `google.com` using the `readme` app, which is presumably due to the overhead of the dioxus event handlers. So we'll want to look at that at some point.

nicoburns:
@luoxiaozero FYI this broke clicking to focus text inputs due to conflicts. Fixed now, but might be a good idea to get into the habit of rebasing PRs against latest main.

Issue #181, opened 2025-01-20
Add `ureq` backed to `blitz-net`


nicoburns:
The motivation for this is primarily as a binary size optimisation for apps that want networking but do not need 100s of connections and do not otherwise need `tokio` or an async executor in their stack.

The implementation should:

- [ ] Be part of the existing `blitz-net` crate with feature flags for each backend
- [ ] Use the new 3.x version of `ureq`
- [ ] Split the `reqwest` and `ureq` implementations in separate files/modules
- [ ] Use a threadpool to allow for multiple concurrent requests
- [ ] Implement the `blitz_net::get_text` method as well as `blitz_net::Provider`

PR #189, opened 2025-02-28
Add `ureq` backend for `blitz-net`

SpikeHD:
Addresses #181 

# TODO

(These are just copied from the above issue :P)
- [x] Be part of the existing blitz-net crate with feature flags for each backend
- [x] Use the new 3.x version of ureq
- [x] Split the reqwest and ureq implementations in separate files/modules
- [x] Use a threadpool to allow for multiple concurrent requests
- [x] Implement the blitz_net::get_text method as well as blitz_net::Provider


SpikeHD:
To test this, set `ureq` as the default feature in `blitz-net`, then change `packages/blitz/src/lib.rs@L36` like so:
```diff
- let html = rt.block_on(blitz_net::get_text(&url));
+ let html = blitz_net::get_text(&url));
```

SpikeHD:
@nicoburns I know [you'll be gone](https://github.com/DioxusLabs/blitz/pull/188#issuecomment-2691465248), but once you're back, I have a couple architecture questions:
* Is there a better way to handle the two `Provider`s? Right now they are almost the same, except for the fact that `fetch_inner` is `async` on the `reqwest` backend. I'd like to ofc prevent repeated code but my little brain can't conceptualize how that might look
* How should the other packages be modified to not use `tokio` (in the case someone is gonna use the `ureq` backend to avoid an async runtime, for example)? From what I can tell, three packages use it, `blitz`, `blitz-shell`, and `dioxus-native`. Or is that out of scope for this PR (I would be very okay with this :P)?

"Allow edits by maintainers" should be on if you feel inspired haha

nicoburns (changes requested):
Sorry for taking so long to review this. I've requested a couple of specific changes.

In regard to your questions:
- I think you should be able to unify the interface of the backends by using a callback-based interface which should work for both async and threadpool+channel based backends.
- Removing `tokio` from other crates has now happened :)

I think I'm probably going to try to land https://github.com/DioxusLabs/blitz/pull/188 before this as I suspect it will be easier to rework this one than that one.

PR #199, opened 2025-03-26
Handle single quotes in WPT reftest regex

valadaptive:
Ran into this while testing https://github.com/linebender/parley/pull/315. This was causing a lot of reftests to be skipped.

Most of these tests still don't pass even with that Parley change, but it seems to be because the `width` attribute of the test `<div>` in the tests isn't being respected.

### DIFF
--- a/apps/wpt/src/main.rs
+++ b/apps/wpt/src/main.rs
@@ -371,7 +371,7 @@ fn main() {
                         ColorScheme::Light,
                     );
                     let net_provider = Arc::new(WptNetProvider::new(&wpt_dir));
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
This regex isn't perfectly correct as it should really allow single quotes within double quotes and vice-versa. But we'll give this a go.

I was wondering why those tests weren't being detected as "ref" tests.

The reason they're failing (aside from any defects there may be with our implementation) is that the width is being set with JavaScript (presumably to work around discrepancies in font metrics between the system fonts of the system the test is being run on) which Blitz doesn't currently support.

PR #200, opened 2025-03-27
Support img src attribute for dioxus-native

otgerrogla:
Currently setting a value to an `img` tag `src` attribute in dioxus does nothing when using the dioxus-native renderer. Like:
````
img {
  src: "some_image_filepath_or_url"
}
````
With this small change, when a change on this attribute is reflected into the real dom, it also triggers a resource load for the image so the image is actually displayed (provided there is a `net_provider` configured).


nicoburns (approved):

nicoburns:
We need to move this functionality into `blitz-dom`. But this seems fine in the meantime :)

You'll need to format the code.

PR #201, opened 2025-03-31
Support autofocus attribute for dioxus-native

otgerrogla:
Add support for the [global "autofocus" attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) to dioxus native. It behaves like in the browsers: it is just taken into account when the element is first added to the DOM, and changing its value afterwards will do nothing.

As of now it only works if it is set to a string value on dioxus, such as `autofocus: "true"`, but not `autofocus: true`, due to non-string arguments being currently ignored for native


nicoburns (changes requested):
Please can you:
- Remove the `PostMutationEffects` struct
- Make `node_to_autofocus` a field of `MutationWriter` instead
- Move the application of the post mutation effects into the `Drop` impl for `MutationWriter` (where is already one effect)

I'm also a little concerned about the perf impact here. But I can't think of a better way to implement this so I won't make that blocking.

You'll also need to fix the clippy lints.

otgerrogla:
Yeah, there will be a perf impact and probably most users won't even need this attribute, as it is somewhat niche.
Maybe it would be better to add it behind a create feature flag for dioxus-native?

nicoburns:
> Yeah, there will be a perf impact and probably most users won't even need this attribute, as it is somewhat niche. Maybe it would be better to add it behind a create feature flag for dioxus-native?

An `autofocus` feature flag for this makes sense to me. And fits with our general Blitz philosophy of being configurable and allowing users to "pay for what they use".

nicoburns (approved):

PR #202, opened 2025-04-01
fix: Fixed the actual color displayed when border-color is transparent

luoxiaozero:
When using the following styles, the `border-color` should appear `red` instead of `transparent`.

```css
#transparent {
    border: 20px solid transparent;
    background-color: red;
}
```

### DIFF
--- a/examples/assets/border.html
+++ b/examples/assets/border.html
@@ -1,0 +1,116 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Document</title>
+        <style>
+            .colorful {
+                border-right-color: #000;
+                border-left-color: #ff0;
+                border-top-color: #f01;
+                border-bottom-color: #0f0;
+            }
+            #a {
+                height: 300px;
+                background-color: gray;
+                border: 1px solid black;
+                border-radius: 50px 20px;
+                border-top-color: red;
+                padding: 20px;
+                margin: 20px;
+                border-radius: 10% 30% 50% 70%;
+                border-left: 4px solid #000;
+                border-top: 10px solid #ff0;
+                border-right: 3px solid #f01;
+                border-bottom: 9px solid #0f0;
+                box-shadow: 10px 10px gray;
+            }
+
+            #b {
+                border: 20px solid black;
+                background-color: red;
+                border-radius: 10px;
+                border-top-width: 32px;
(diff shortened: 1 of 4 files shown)


nicoburns (changes requested):
This really ought to be implemented as:
- Not drawing the border at all
- Having the background show through (I suspect we're currently incorrectly clipping the background to the padding box)

This implementation works for solid colours. But is going to fail for background images.

nicoburns (changes requested):

luoxiaozero:
Implement the following functions:

* [x] background-clip
* [x] background-origin
* [x] background-size
* [x] background-origin

To prevent this PR from becoming too large, I did not implement `background-repeat`.  Currently it behaves the same as `no-repeat`.

nicoburns (approved):
I was struggling to follow the logic for some of this PR, so I tried to run the WPT `css/css-backgrounds` tests to see how this PR effected the results of those. It turned out that our test runner was crashing on a lot of these tests, so I've spent a few hours fixing that (https://github.com/DioxusLabs/blitz/pull/206)

Results with the fixed test runner are that this makes around 60 more tests pass compared to `main` (so it seems good to merge!). However, it's worth noting that there are still quite a few background tests not passing (including for properties like `background-size` which we do support).

To run tests: `cargo run -rp wpt css/css-backgrounds`.

Also, if you're planning to implement background-repeat, then note that Vello has a built-in ability to repeat raster images which we probably ought to use.

nicoburns:
@luoxiaozero I had to revert this as it was quite badly broken on displays with a non-1.0 scale factor

luoxiaozero:
Does the damage on a non-1.0 scale factor display look the same as in the comparison image below? If not, can you provide a comparison image?

| scale_factor: 1.25 | scale_factor: 1 |
| -- | -- |
| fc0c5df7-5ab3-42af-aab6-ec06041b4e72 |518d9015-ae8b-4493-8979-0643f84194db|

nicoburns:
The issue appears to be the backgrounds being too large by a factor of the scale factor.

A couple of comparison images (scale factor 2):

`cargo run -rp todomvc` (main): 

Screenshot 2025-04-10 at 23 31 50

`cargo run -rp todomvc` (this PR):

Screenshot 2025-04-10 at 23 32 30

Footer of `cargo run -rp readme alistapart.com` (main):

Screenshot 2025-04-10 at 23 33 10

Footer of `cargo run -rp readme alistapart.com` (this PR):

Screenshot 2025-04-10 at 23 32 52

PR #210, opened 2025-05-04
feat: Support for background-repeat properties for images

lizidev:
Running `cargo run -rp wpt css/css-backgrounds`, 22 more passes than main (173 -> 195).


nicoburns:
Going to merge this as I think it's an improvement. But in future I would appreciate having "move to separate file" and "make substantive changes" in separate PRs, because this one doesn't have a useful diff at all.

PR #214, opened 2025-05-20
fix: Fixed some rendering behaviors of box-shadow

lizidev:
* Fix shadow clipping.
* Fixed shadow rendering order.
* Support for the spread property of shadows.

### DIFF
--- a/examples/box_shadow.rs
+++ b/examples/box_shadow.rs
@@ -1,0 +1,48 @@
+use dioxus::prelude::*;
+
+fn main() {
+    dioxus_native::launch(app);
+}
+
+fn app() -> Element {
+    rsx! {
+        div {
+            style { {CSS} }
+            div {
+                id: "box-shadow-1",
+                class: "box-shadow",
+            }
+            div {
+                id: "box-shadow-2",
+                class: "box-shadow",
+            }
+            div {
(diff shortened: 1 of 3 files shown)


lizidev:
`error: this if statement can be collapsed` error in cargo clippy, I don't know which is the better solution:

* Add `rust-version` to all crates, [related issues](https://github.com/rust-lang/rust-clippy/issues/14850). 
* Add `#[allow(clippy::collapsible_if)]` to all errors.

nicoburns:
@lizidev
- I've updated CI to use stable rust rather than nightly rust for clippy.
- I've updated `color` on `main` so you shouldn't need that here
- Do you have screenshots, tests, or sample code which demonstrates the rendering features you have fixed here?

lizidev:
Using the example I added, the order of the list below corresponds to the order in the screenshots:

* Fix shadow clipping.
* Fixed shadow rendering order.
* Support for the spread property of shadows.

| main branch |  This PR |
|-| -|
|2561747752446_ pic|2571747752542_ pic|

nicoburns:
This seems to fix the shadow rendering for the `todomvc` example - nice!
(see: the "stacked boxes" effect at bottom of the central card)

Screenshot 2025-05-20 at 17 01 00

PR #239, opened 2025-07-20
<input type="file">

kokoISnoTarget:
This adds:
- `<input type="file">`
- `multipart/form-data` form submission
- File picker integration in `blitz-shell`


packages/blitz-dom/src/form.rs
```diff
@@ -446,46 +460,226 @@ impl Display for RequestContentType {
...
+}
+
+/// Creates a part of a multipart/form-data body
+fn create_part<W: std::io::Write>(w: &mut W, name: &str, value: &EntryValue, boundary: &str) {
+    //TODO: Either do this by removing from CONTROLS or encode this without the percent encoding crate
+    const MINIMAL_ENCODE_SET: percent_encoding::AsciiSet =
+        unsafe { std::mem::transmute::<[u32; 4], percent_encoding::AsciiSet>([0u32; 4]) }
+            .add(b'\n')
+            .add(b'\r')
+            .add(b'"');
```
nicoburns: I think we should avoid the transmute here. Deriving from `CONTROLS` is probably the best option? We could probbably just add to `CONTROLS`?

packages/blitz-dom/src/mutator.rs
```diff
@@ -620,6 +621,45 @@ impl<'doc> DocumentMutator<'doc> {
             let value = value.to_string();
             let id = self.create_text_node(&value);
             self.append_children(target_id, &[id]);
+        } else if let ("input", Some("file")) = (tagname, type_attr) {
+            let button_id = self.create_element(
+                QualName {
+                    prefix: None,
+                    ns: ns!(html),
+                    local: local_name!("button"),
+                },
```
nicoburns: Optional: might be worth creating a `macro_rules` macro for creating these?

packages/blitz-dom/src/node/element.rs
```diff
@@ -261,6 +280,16 @@ impl ElementData {
     pub fn take_inline_layout(&mut self) -> Option<Box<TextLayout>> {
         std::mem::take(&mut self.inline_layout_data)
     }
+
+    pub fn is_submit_button(&self) -> bool {
+        let is_button = self.name.local == local_name!("button");
```
nicoburns: This should probably return early if it's not a button.

packages/blitz-shell/src/lib.rs
```diff
@@ -144,4 +157,33 @@ impl ShellProvider for BlitzShellProvider {
...
+            target_os = "windows",
+            target_os = "macos",
+            target_os = "linux",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "netbsd",
+            target_os = "openbsd"
+        )
+    ))]
+    fn open_file_dialog(
```
nicoburns, line 173: I really like how you've put the file input behind the `ShellProvider` trait.

packages/blitz-dom/src/form.rs
kokoISnoTarget: Would you allow this until the next release of the percent_encoding crate, which should have an empty ASCII set?

nicoburns:
> > Can we feature flag file inputs in `blitz-dom`?
> Just user facing stuff or everything (like SpecialElementData::FileInput)?

Everything. Idea is that you don't pay the cost of the feature if it's not enabled.

> > Can you add an example somewhere? (and/or suggest a URL I can test the feature with)
> 
> Is a python test server fine? If not, you can always post against something like https://httpbin.org/forms/post using the readme app

Perhaps just add a file input to the `form` example? Even if it doesn't post, at least we could test the UI.

> > I think we ought to define a typed interface in blitz-traits that allows key-value pairs to be sent to blitz-net that would then encode them
> 
> I would be happy to work on that in the next few days

🎉

packages/blitz-dom/src/form.rs
nicoburns: Can we just also encode control characters until then? Would we not want to encode control characters anyway (in practice it probably doesn't make much difference as file names are unlikely to include them?)

Cargo.toml
```diff
@@ -100,7 +101,7 @@ url = "2.5.0"
 http = "1.1.0"
 data-url = "0.3.1"
 tokio = "1.42"
-reqwest = "0.12"
+reqwest = { version = "0.12", features = ["multipart", "stream"] }
```
nicoburns: These features should be:
- Enabled in the specific crate they are used
- Enabled conditionally based on a feature flag.

examples/form.rs
```diff
@@ -92,6 +92,28 @@ fn app() -> Element {
...
+                    label {
+                        r#for: "file1",
+                        "File Select Single",
+                    }
+                }
+                div {
+                    input {
+                        type: "file",
+                        name: "single_multi",
+                        id: "file2",
```
nicoburns, line 110: Should this one have a `multiple` attribute?

packages/blitz-dom/src/events/mouse.rs
```diff
@@ -153,65 +153,106 @@ pub(crate) fn handle_click<F: FnMut(DomEvent)>(
...
+                    1 => files
+                        .first()
+                        .unwrap()
+                        .file_name()
+                        .unwrap_or_default()
+                        .to_str()
+                        .expect("Filename is not UTF-8")
+                        .to_string(),
+                    x => format!("{x} Files Selected"),
+                };
```
nicoburns: Do we need to require UTF-8? Could do `to_utf8_lossy`? (probably better than crashing?)

examples/form.rs
kokoISnoTarget: oops

PR #244, opened 2025-08-03
Use `from_peniko_image` to convert `peniko::Image` to `Pixmap`

sagudev:
This will also respect opacity in `peniko::Image`.


packages/anyrender_vello_cpu/src/scene.rs
```diff
@@ -10,75 +10,30 @@ fn brush_ref_to_paint_type<'a>(brush_ref: BrushRef<'a>) -> PaintType {
...
-                PremulRgba8 {
-                    r: premultiply(d[0], alpha),
-                    g: premultiply(d[1], alpha),
-                    b: premultiply(d[2], alpha),
-                    a: d[3],
-                }
-            }
-        })
-        .collect()
+    map.entry(id).or_insert_with(|| vello_cpu::Image::from_peniko_image(image)).clone()
```
sagudev: `vello_cpu::Image` has `Arc<Pixmap>` inside

nicoburns (approved):
Thanks!

PR #247, opened 2025-08-13
Incremental/parallel tree construction

nicoburns:
# Objective

Greatly improve performance by:
- Doing construction / layout / paint incrementally where possible.
- Doing some parts of the construction in parallel

### Parley

- [x] Invalidate all parley layouts when a new font is loaded
- [x] Invalidate all parley layouts when scale factor changes
- [x] Make underline/strikethrough/textcolor updates work without rebuilding parley layout
- [x] Reuse `parley::Layout` when reconstructing (use `build_into`)
- [x] Cache intrinsic sizes + compute in pconstruct phase (if no children)

### Bugs
- [x] Missing text (servo.org, wikipedia.org)
- [x] Images only load in on resize
- [x] Pseudo-element alignment bug https://github.com/DioxusLabs/blitz/pull/247#issuecomment-3235111617

### Optimisations

- [x] Don't request redraw if scroll event has no effect (try to scoll past end of doc)
- [ ] ~~Optimise style diffing (ptr comparisons)~~ moved to issue
- [ ] ~~Limit damage to formatting context where appropriate~~ moved to issue
- [ ] ~~Make use of REPAINT damage (don't repaint if no damage)~~ moved to issue


nicoburns:
@Aeledfyr That is really bizarre behaviour. Thanks for the reproduction! I think I've managed to track down the issue.

This is a bit technical, but what was happening was that the "measure content width" stages of subsequent layout runs were "clobbering" the layout in order to measure the text (but were unable to correctly compute alignment as they didn't yet have the width to align to available at that stage). Normally the later "final layout" stage would overwrite that, but in this case that stage was hitting the cache and not running at all.

The fix I've pushed is a bit of hack which runs the "measure content width" stage on a clone of the layout. I would have expected this to be slow, but in practice it doesn't seem to make much difference. So I've put it in for now, and we can clean it up later.

Issue #241, opened 2025-07-31
Unreset background colour is off-white on CPU renderer


Jamedjo:
The body background color is white when using the GPU renderer, but off-white for the CPU renderer.

Adding `body { background-color: white;}` fixes the discrepancy between renderers.

As a nice bonus the image is looking much sharper with the CPU renderer and the blue/green fix was appreciated ❤️ 

<img width="1904" height="1008" alt="Image" src="https://github.com/user-attachments/assets/0a43f5a7-8e2d-40d9-b702-8c1f1f246e01" />

MarcAntoine-Arnaud:
Isn't it not easiest to update that line: https://github.com/DioxusLabs/blitz/blob/main/packages/anyrender_vello/src/image_renderer.rs#L115
from: 
`base_color: vello::peniko::Color::WHITE,`
with:
`base_color: vello::peniko::Color::TRANSPARENT,`

T-256:
solved by #248 ?

PR #248, opened 2025-08-21
fix: transparent background colour for GPU - same as for CPU

MarcAntoine-Arnaud:
issue #241


nicoburns:
@MarcAntoine-Arnaud Seems reasonable. Have you verified that this works?

MarcAntoine-Arnaud:
@nicoburns yes I have checked and it works.

nicoburns:
Excellent - let's get this merged then :)

Issue #211, opened 2025-05-06
Support the `clip-path` property


nicoburns:
We should support the `clip-path` property:

- https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path
- https://drafts.fxtf.org/css-masking/#the-clip-path

Our underlying rendering engine(s) support arbitrary paths as clips so this mostly be a matter of:
- Translating Stylo's [ClipPath](https://docs.rs/stylo/latest/style/values/computed/basic_shape/type.ClipPath.html) into a Kurbo [impl Shape](https://docs.rs/kurbo/latest/kurbo/trait.Shape.html)
- Some logic to resolve the path against the specific [geometry-box](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path#geometry-box)

Tricky cases could be left out of the initial implementation.

scriptforfivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

nicoburns:
Implemented in https://github.com/DioxusLabs/blitz/commit/7ecd70b7cc4c53be826dae572ab356f1947b4c30

PR #250, opened 2025-08-30
WIP: Support the clip-path property #211

gautam1168:
[Issue](https://github.com/DioxusLabs/blitz/issues/211#issuecomment-3216565893)

The code is completely work in progress and not intended to be considered for merging yet. Im opening this pull request so that a maintainer can check that I am not going off in a direction that will be unmergeable. Mainly, I want to show how much the code change would be. It felt like I might be writing too much for a first issue. 

To run this just run the `outline` example. It will show a screen with some clip paths:
Screenshot 2025-08-30 at 3 55 23 PM
Screenshot 2025-08-30 at 3 55 33 PM

### DIFF
--- a/packages/blitz-paint/src/render.rs
+++ b/packages/blitz-paint/src/render.rs
@@ -368,7 +368,124 @@ impl BlitzDomPainter<'_> {
                         };
                         &cx.frame.circle_path(center, radius)
                     },
-                   _ => &cx.frame.padding_box_path()
+                    GenericBasicShape::Ellipse(e) => {
+                        let center: Point = match e.position {
+                            GenericPositionOrAuto::Position(pos) => {
+                                let hor = match pos.horizontal.unpack() {
+                                    Unpacked::Length(l) => l.px() as f64,
+                                    Unpacked::Percentage(p) => box_width * p.to_percentage() as f64 / 2.0,
+                                    Unpacked::Calc(calc) => calc.resolve(CSSPixelLength::new(box_width as f32)).px() as f64 / 2.0,
+                                };
+
+                                let vert = match pos.vertical.unpack() {
+                                    Unpacked::Length(l) => l.px() as f64,
+                                    Unpacked::Percentage(p) => box_height * p.to_percentage() as f64 / 2.0,
+                                    Unpacked::Calc(calc) => calc.resolve(CSSPixelLength::new(box_height as f32)).px() as f64 / 2.0,
+                                };
+
+                                Point {
+                                    x: hor + cx.frame.padding_box.origin().x as f64,
+                                    y: vert + cx.frame.padding_box.origin().y as f64,
+                                }
+                            },
+                            GenericPositionOrAuto::Auto => {
+                                let center_x = box_width / 2.0;
+                                let center_y = box_height / 2.0;
+                                Point {
+                                    x: center_x as f64,
+                                    y: center_y as f64,
+                                }
+                            }
+                        };
+                        let radius_x = match e.semiaxis_x {
+                            ShapeRadius::Length(radius) => {
+                                match radius.0.unpack() {
+                                    Unpacked::Length(l) => l.px() as f64,
+                                    Unpacked::Percentage(p) => box_width * p.to_percentage() as f64 / 2.0,
+                                    Unpacked::Calc(calc) => calc.resolve(CSSPixelLength::new(box_width as f32)).px() as f64 / 2.0,
+                                }
+                            },
+                            _ => box_width as f64 / 2.0,
+                        };
+                        let radius_y = match e.semiaxis_y {
+                            ShapeRadius::Length(radius) => {
+                                match radius.0.unpack() {
+                                    Unpacked::Length(l) => l.px() as f64,
+                                    Unpacked::Percentage(p) => box_height * p.to_percentage() as f64 / 2.0,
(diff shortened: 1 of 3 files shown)


packages/blitz-paint/src/render.rs
```diff
@@ -224,7 +426,341 @@ impl BlitzDomPainter<'_> {
...
+
+                                let vert = match pos.vertical.unpack() {
+                                    Unpacked::Length(l) => l.px() as f64,
+                                    Unpacked::Percentage(p) => {
+                                        box_height * p.to_percentage() as f64 / 2.0
+                                    },
+                                    Unpacked::Calc(calc) => {
+                                        calc.resolve(CSSPixelLength::new(box_height as f32)).px() as f64 / 2.0
+                                    }
+                                };
```
nicoburns, line 533: I think this can just be `pos.vertical.resolve(CSSPixelLength::new(box_height as f32 / 2.0))` (see https://docs.rs/stylo/latest/style/values/computed/length_percentage/struct.LengthPercentage.html#method.resolve)

nicoburns: But what is the `/ 2.0` for? Are you missing a scale factor conversion?

packages/blitz-paint/src/render.rs
```diff
@@ -224,7 +426,341 @@ impl BlitzDomPainter<'_> {
...
+                &cx.frame.padding_box_path()
+            },
+            GenericClipPath::Box(geometry_box) => {
+                match geometry_box {
+                    ShapeGeometryBox::ShapeBox(b) => {
+                        match b {
+                            ShapeBox::BorderBox => &cx.frame.border_box_path(),
+                            ShapeBox::PaddingBox => &cx.frame.padding_box_path(),
+                            ShapeBox::ContentBox => &cx.frame.content_box_path(),
+                            ShapeBox::MarginBox => &cx.frame.border_box_path(),
```
nicoburns, line 457: This should be margin box. If you don't want to implement yet, then a TODO comment would be fine.

packages/blitz-paint/src/render.rs
```diff
@@ -224,7 +426,341 @@ impl BlitzDomPainter<'_> {
...
+                                cx.frame.padding_box.height(),
+                            ),
+                            ShapeBox::ContentBox => (
+                                cx.frame.content_box.width(),
+                                cx.frame.content_box.height(),
+                            ),
+                            ShapeBox::MarginBox => (
+                                cx.frame.border_box.width(),
+                                cx.frame.border_box.height(),
+                            ),
```
nicoburns, line 484: Same here, can we leave a comment that margin box is TODO (just makes it clear to anyone reading the code why this was set to border box)

packages/blitz-paint/src/render.rs
```diff
@@ -224,7 +426,341 @@ impl BlitzDomPainter<'_> {
...
+                            Unpacked::Percentage(p) => box_width * p.to_percentage() as f64,
+                            Unpacked::Calc(calc) => calc.resolve(CSSPixelLength::new(box_width as f32)).px() as f64 / 2.0,
+                        };
+                        let y1 = match r.rect.3.unpack() {
+                            Unpacked::Length(l) => l.px() as f64,
+                            Unpacked::Percentage(p) => box_height * p.to_percentage() as f64,
+                            Unpacked::Calc(calc) => calc.resolve(CSSPixelLength::new(box_height as f32)).px() as f64 / 2.0,
+                        };
+                        // Kurbo doesn't support rounded rect with elliptical radii,
+                        // so we ignore the y axes right now.
```
nicoburns, line 624: Fine for now.

nicoburns (changes requested):
@gautam1168 Firstly, no worries on the delay. This is open source, you can do thing when you like.

A couple of notes:

- Kurbo has `SvgArc`, which seems to have the same parameterization as Stylo. And then a built-in function to convert that to it's regular `Arc` representation.
- It also have this [BezPath::from_svg](https://docs.rs/kurbo/latest/kurbo/struct.BezPath.html#method.from_svg) function to parse an string SVG path into a `BezPath`. I don't think you can use that directly, but I think you could base your conversion code on it's source (https://docs.rs/kurbo/latest/src/kurbo/svg.rs.html#105)

packages/blitz-paint/src/render.rs
gautam1168: Yes this is one of those things I would look into deeper. But the (/2.0) is there because I was thinking the circle's radius would be half the box size. But I haven't thought too hard about this yet. I have also ignored the `by_to` parameter for all the cases right now just to get a compilable/runnable first draft.

I will checkout the svgArc and replace Arc with SvgArc. Hopefully that will simplify the code a lot and get rid of a bunch of convertions.

gautam1168:
Hey thanks for taking the time to go over this. Ill work on the things you have mentioned here.

gautam1168:
Hey @nicoburns I have suddenly become aware of very serious health issues. These have been bothering me for a while but I have finally reached the breaking point it seems. I may not be able to complete this PR in any reasonable timeframe. I will work on it if my health permits but if you want to close this one or let someone else work on it that would be fine. 

Sorry, for the inconvenience caused here. I will make up for it if I get healthy again. 

Best regards
gaurav

nicoburns:
Implemented in https://github.com/DioxusLabs/blitz/commit/7ecd70b7cc4c53be826dae572ab356f1947b4c30

PR #269, opened 2025-10-02
Update to wgpu 26, vello 0.6, vello_cpu 0.0.3, and parley pre-0.6

rydb:
This pr updates blitz to use wgpu 26 and related dependencies to accommodate this update.

I've left a few questions marked with `//rydb:`, but otherwise, once vello and parley are updated, I can update those plus add any additional changes to this.

### DIFF
--- a/examples/wgpu_texture/src/demo_renderer.rs
+++ b/examples/wgpu_texture/src/demo_renderer.rs
@@ -210,6 +210,7 @@ impl ActiveDemoRenderer {
                         load: wgpu::LoadOp::Clear(wgpu::Color::GREEN),
                         store: wgpu::StoreOp::Store,
                     },
+                    depth_slice: None,
                 })],
                 depth_stencil_attachment: None,
                 timestamp_writes: None,
--- a/examples/wgpu_texture/src/demo_renderer.rs
+++ b/examples/wgpu_texture/src/demo_renderer.rs
@@ -170,11 +170,13 @@ impl ActiveDemoRenderer {
         start_time: &Instant,
     ) -> Option<TextureHandle> {
         // If "next texture" size doesn't match specified size then unregister and drop texture
-        if let Some(next) = &self.next_texture {
-            if next.texture.width() != width || next.texture.height() != height {
-                ctx.unregister_texture(next.handle.clone());
-                self.next_texture = None;
-            }
+        if self
+            .next_texture
+            .as_ref()
+            .is_some_and(|tex| tex.texture.width() != width || tex.texture.height() != height)
+        {
+            let handle = self.next_texture.take().unwrap().handle;
+            ctx.unregister_texture(handle);
         }
 
         // If there is no "next texture" then create one and register it.
--- a/examples/wgpu_texture/src/demo_renderer.rs
+++ b/examples/wgpu_texture/src/demo_renderer.rs
@@ -172,7 +172,7 @@ impl ActiveDemoRenderer {
         // If "next texture" size doesn't match specified size then unregister and drop texture
         if let Some(next) = &self.next_texture {
             if next.texture.width() != width || next.texture.height() != height {
-                ctx.unregister_texture(next.handle);
+                ctx.unregister_texture(next.handle.clone());
                 self.next_texture = None;
             }
         }
@@ -189,7 +189,7 @@ impl ActiveDemoRenderer {
         };
 
         let next_texture = &texture_and_handle.texture;
-        let next_texture_handle = texture_and_handle.handle;
+        let next_texture_handle = texture_and_handle.handle.clone();
 
         let elapsed: f32 = start_time.elapsed().as_millis() as f32 / 500.;
         let [light_red, light_green, light_blue] = light;
--- a/packages/anyrender_vello/src/custom_paint_source.rs
+++ b/packages/anyrender_vello/src/custom_paint_source.rs
@@ -1,7 +1,7 @@
 use crate::wgpu_context::DeviceHandle;
 use peniko::ImageData;
 use vello::Renderer as VelloRenderer;
-use wgpu::{Instance, TexelCopyTextureInfoBase, Texture};
+use wgpu::{Instance, Texture};
 
 pub trait CustomPaintSource: 'static {
     fn resume(&mut self, instance: &Instance, device_handle: &DeviceHandle);
@@ -28,20 +28,10 @@ impl CustomPaintCtx<'_> {
     }
 
     pub fn register_texture(&mut self, texture: Texture) -> TextureHandle {
-        let dummy_image = self.renderer.register_texture(texture.clone());
-
-        let base = TexelCopyTextureInfoBase {
-            texture: texture,
-            mip_level: 0,
-            origin: wgpu::Origin3d::ZERO,
-            aspect: wgpu::TextureAspect::All,
-        };
-        self.renderer.override_image(&dummy_image, Some(base));
-
-        TextureHandle(dummy_image)
+        TextureHandle(self.renderer.register_texture(texture))
     }
 
     pub fn unregister_texture(&mut self, handle: TextureHandle) {
-        self.renderer.override_image(&handle.0, None);
+        self.renderer.unregister_texture(handle.0);
     }
 }
--- a/packages/anyrender_vello/src/custom_paint_source.rs
+++ b/packages/anyrender_vello/src/custom_paint_source.rs
@@ -1,7 +1,5 @@
-use std::sync::Arc;
-
 use crate::wgpu_context::DeviceHandle;
-use peniko::{Blob, ImageBrush, ImageData};
+use peniko::ImageData;
 use vello::Renderer as VelloRenderer;
 use wgpu::{Instance, TexelCopyTextureInfoBase, Texture};
 
@@ -21,18 +19,8 @@ pub struct CustomPaintCtx<'r> {
     pub(crate) renderer: &'r mut VelloRenderer,
 }
 
-#[derive(Copy, Clone, PartialEq, Hash)]
-pub struct TextureHandle {
-    pub(crate) id: u64,
-    pub(crate) width: u32,
-    pub(crate) height: u32,
-}
-
-impl TextureHandle {
-    pub(crate) fn dummy_image(&self) -> ImageBrush {
-        dummy_image(Some(self.id), self.width, self.height)
-    }
-}
+#[derive(Clone, PartialEq)]
+pub struct TextureHandle(pub ImageData);
 
 impl CustomPaintCtx<'_> {
     pub(crate) fn new<'a>(renderer: &'a mut VelloRenderer) -> CustomPaintCtx<'a> {
@@ -40,41 +28,20 @@ impl CustomPaintCtx<'_> {
     }
 
     pub fn register_texture(&mut self, texture: Texture) -> TextureHandle {
-        let dummy_image = dummy_image(None, texture.width(), texture.height());
-        let handle = TextureHandle {
-            id: dummy_image.image.data.id(),
-            width: texture.width(),
-            height: texture.height(),
-        };
+        let dummy_image = self.renderer.register_texture(texture.clone());
+
         let base = TexelCopyTextureInfoBase {
-            texture,
+            texture: texture,
             mip_level: 0,
             origin: wgpu::Origin3d::ZERO,
             aspect: wgpu::TextureAspect::All,
         };
-        self.renderer.override_image(&dummy_image.image, Some(base));
+        self.renderer.override_image(&dummy_image, Some(base));
 
-        handle
+        TextureHandle(dummy_image)
     }
 
     pub fn unregister_texture(&mut self, handle: TextureHandle) {
-        let dummy_image = dummy_image(Some(handle.id), handle.width, handle.height);
-        self.renderer.override_image(&dummy_image.image, None);
+        self.renderer.override_image(&handle.0, None);
     }
-}
-
-// Everything except blob id, width, and height is ignored
-fn dummy_image(id: Option<u64>, width: u32, height: u32) -> ImageBrush {
-    let blob = match id {
-        Some(id) => Blob::from_raw_parts(Arc::new([]), id),
-        None => Blob::new(Arc::new([])),
-    };
-
-    ImageBrush::new(ImageData {
-        data: blob,
-        width,
-        height,
-        format: vello::peniko::ImageFormat::Rgba8,
-        alpha_type: vello::peniko::ImageAlphaType::Alpha,
-    })
 }
--- a/packages/anyrender_vello/src/custom_paint_source.rs
+++ b/packages/anyrender_vello/src/custom_paint_source.rs
@@ -1,9 +1,8 @@
 use std::sync::Arc;
 
 use crate::wgpu_context::DeviceHandle;
-use peniko::Blob;
+use peniko::{Blob, ImageBrush, ImageData};
 use vello::Renderer as VelloRenderer;
-use vello::peniko::Image;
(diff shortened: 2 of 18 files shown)


packages/blitz-paint/src/layers.rs
```diff
@@ -47,8 +47,8 @@ pub(crate) fn maybe_push_layer(
     if !layers_available {
         return false;
     }
-
     let blend_mode = if opacity == 1.0 {
+        // rydb: `Mix::Clip` deprecation warning recommends using `push_clip_layer`, but peniko doesn't have that function?
```
nicoburns: `push_clip_layer` is on `vello::Scene`. But it's fine to leave this for now. Vello won't be removing `Mix::Clip` until the *next* version.

packages/blitz-paint/src/render.rs
```diff
@@ -395,6 +400,11 @@ struct ElementCx<'a> {
     devtools: &'a DevtoolSettings,
 }
 
+///TODO: If BoundingBox implements `Shape`, remove this. This exists to side-step it not currently doing that.
```
nicoburns: `BoundingBox` is never going to implement `Shape`. So maybe just make this comment describe what the function does.

packages/blitz-paint/src/render.rs
```diff
@@ -395,6 +400,11 @@ struct ElementCx<'a> {
...
+fn convert_rect(rect: &parley::BoundingBox) -> peniko::kurbo::Rect {
+    peniko::kurbo::Rect::new(rect.x0, rect.y0, rect.x1, rect.y1)
```
nicoburns: No need to reference `kurbo` via `peniko` here. `blitz-paint` depends on `kurbo` directly.

packages/anyrender_vello_cpu/src/scene.rs
```diff
@@ -1,39 +1,68 @@
...
+            PaintType::Image(brush)
+        }
     }
+    // match brush_ref {
+    //     BrushRef::Solid(alpha_color) => PaintType::Solid(alpha_color),
+    //     BrushRef::Gradient(gradient) => PaintType::Gradient(gradient.clone()),
+    //     BrushRef::Image(image) => {
+    //         PaintType::Image(vello_cpu::Image::from_peniko_image(&image))
+    //     },
+    // }
```
nicoburns: Commented code (please remove)

packages/anyrender_vello_cpu/src/scene.rs
```diff
@@ -1,39 +1,68 @@
...
-        BrushRef::Solid(alpha_color) => PaintType::Solid(alpha_color),
-        BrushRef::Gradient(gradient) => PaintType::Gradient(gradient.clone()),
-        BrushRef::Image(image) => PaintType::Image(vello_cpu::Image::from_peniko_image(image)),
+        Brush::Solid(alpha_color) => PaintType::Solid(alpha_color),
+        Brush::Gradient(gradient) => PaintType::Gradient(gradient.clone()),
+        Brush::Image(image) => {
+            let image_source = vello_cpu::ImageSource::from_peniko_image_data(&image.image);
+            let brush = image_source_to_image_brush(image_source);
+            PaintType::Image(brush)
+        }
```
nicoburns: I think this should be something like:

```rust
Brush::Image(image) => {
     let image_source = vello_cpu::ImageSource::from_peniko_image_data(&image.image);
     PaintType::Image(ImageBrush {
           image: image_source,
           sampler: image.sampler,
     })
}
```

We need to use the sampler parameters from the input, and I don't think the additional helper function is buying us anything here.

packages/anyrender_vello_cpu/src/scene.rs
```diff
@@ -1,39 +1,68 @@
...
+    let id = image.image.data.id();
     map.entry(id)
-        .or_insert_with(|| vello_cpu::Image::from_peniko_image(image))
+        .or_insert_with(|| {
+            let image_source = vello_cpu::ImageSource::from_peniko_image_data(&image.image);
+            let brush = image_source_to_image_brush(image_source);
+            brush
+        })
         .clone()
 }
```
nicoburns: This function should only cache `ImageData` not the whole `ImageBrush.

packages/anyrender_vello/src/image_renderer.rs
```diff
@@ -152,10 +152,23 @@ impl VelloImageRenderer {
...
+                None => panic!("channel closed"),
+            };
+            match result {
+                Ok(_) => {}
+                Err(_err) => {
+                    // rydb: Should this be an panic? There is no logging crate to make this less severe.
+                    // panic!("channel buffer async error: {:#}", err)
+                }
+            }
         }
```
nicoburns, line 162: The extra error handling here is pointless. We know the channel won't have closed because we create it just above. Just unwrap.

packages/anyrender_vello/src/custom_paint_source.rs
```diff
@@ -53,32 +52,29 @@ impl CustomPaintCtx<'_> {
             origin: wgpu::Origin3d::ZERO,
             aspect: wgpu::TextureAspect::All,
         };
-        self.renderer.override_image(&dummy_image, Some(base));
+        self.renderer.override_image(&dummy_image.image, Some(base));
```
nicoburns: This should use the new `register_texture` API:

```rust
let dummy_image = self.renderer.register_texture(texture);
```

And TextureHandle should become a newtype around `peniko::ImageData`.

packages/anyrender_vello/src/custom_paint_source.rs
```diff
@@ -53,32 +52,29 @@ impl CustomPaintCtx<'_> {
...
 
         handle
     }
 
     pub fn unregister_texture(&mut self, handle: TextureHandle) {
         let dummy_image = dummy_image(Some(handle.id), handle.width, handle.height);
-        self.renderer.override_image(&dummy_image, None);
+        self.renderer.override_image(&dummy_image.image, None);
```
nicoburns: This should use the new `unregister_texture` API:

```rust
self.renderer.unregister_texture(&handle.0);
```

(assuming `TextureHandle` wraps an `ImageData`)

packages/anyrender_vello/src/wgpu_context.rs
```diff
@@ -332,10 +365,11 @@ pub fn block_on_wgpu<F: Future>(device: &Device, fut: F) -> F::Output {
...
         match fut.as_mut().poll(&mut context) {
-            std::task::Poll::Pending => {
-                device.poll(wgpu::Maintain::Wait);
-            }
-            std::task::Poll::Ready(item) => break item,
+            std::task::Poll::Pending => match device.poll(wgpu::PollType::Wait) {
+                Ok(_) => continue,
+                Err(err) => return Err(WgpuContextError::PollError(err)),
+            },
+            std::task::Poll::Ready(item) => break Ok(item),
```
nicoburns, line 377: You could probably just `unwrap` here

packages/anyrender_vello_cpu/src/scene.rs
```diff
@@ -1,39 +1,68 @@
 use anyrender::{NormalizedCoord, Paint, PaintScene};
 use kurbo::{Affine, Rect, Shape, Stroke};
-use peniko::{BlendMode, BrushRef, Color, Fill, Font, StyleRef};
-use vello_cpu::{self, PaintType, Pixmap};
+use peniko::{
+    BlendMode, Brush, BrushRef, Color, Fill, FontData, ImageBrush, ImageSampler, StyleRef,
+};
+use vello_cpu::{ImageSource, PaintType, Pixmap};
+// use vello_cpu::{self, ImageSource, PaintType, Pixmap};
```
nicoburns: Commented code (please remove)

packages/blitz-paint/src/layers.rs
nicoburns: We do need to ignore the clippy lints though.

packages/anyrender_vello/src/wgpu_context.rs
rydb: > You could probably just `unwrap` here

Does this need to panic if WGPU fails to poll though?

nicoburns: I'm not sure that there's any useful error handlng that we can do if it does. Do you have any in mind?

rydb: > I'm not sure that there's any useful error handlng that we can do if it does. Do you have any in mind?

Not something specifically, but, since in WGPU 26 polling is  `result`, it might be good for down the line for if blitz adds tracing. `WgpuContextError` could hold the different error types and could push a warning instead of it either being a panic or it causing runtime issues without getting an obvious "why"

nicoburns: I think this is really is a "fatal error" -level error rather than a warning-level error, because we can't meaningfully recover from "GPU is not available". And I also believe the default timeout is 60 seconds, so this error is only likely to occur after the program has already hung for 60 seconds.

rydb: > I think this is really is a "fatal error" -level error rather than a warning-level error, because we can't meaningfully recover from "GPU is not available". And I also believe the default timeout is 60 seconds, so this error is only likely to occur after the program has already hung for 60 seconds.

Is it possible to switch to CPU rendering when that happens though? I.E: GPU poll fails, blitz attempts to switch to CPU rendering if possible, and if that fails, then its fatal?

nicoburns: Fallback to CPU rendering is possible, but I wouldn't expect it to happen here. I'd expect it to happen if/when creation of a WGPU `Device` or `Adapter` fails during setup.

packages/anyrender_vello/src/custom_paint_source.rs
rydb: > This should use the new unregister_texture API:

this change resulted in needing to add a few clones. e.g:
```rust
    pub fn register_texture(&mut self, texture: Texture) -> TextureHandle {
        let dummy_image = self.renderer.register_texture(texture.clone()); //<--- here

        let base = TexelCopyTextureInfoBase {
            texture: texture,
            mip_level: 0,
            origin: wgpu::Origin3d::ZERO,
            aspect: wgpu::TextureAspect::All,
        };
        self.renderer.override_image(&dummy_image, Some(base));

        TextureHandle(dummy_image)
    }
    
```
also in `wgpu_texture/demo_renderer.rs` as well.

nicoburns: See my other comment above about the `register_texture()` method (https://github.com/DioxusLabs/blitz/pull/269#discussion_r2400021730). This should eliminate some clones I think. If there's still the odd clone left, that's ok as `ImageData` and `Texture` are both reference counted, so clones should be relatively cheap.

packages/anyrender_vello/src/custom_paint_source.rs
```diff
@@ -21,60 +19,29 @@ pub struct CustomPaintCtx<'r> {
...
+
         let base = TexelCopyTextureInfoBase {
-            texture,
+            texture: texture,
             mip_level: 0,
             origin: wgpu::Origin3d::ZERO,
             aspect: wgpu::TextureAspect::All,
         };
-        self.renderer.override_image(&dummy_image.image, Some(base));
+        self.renderer.override_image(&dummy_image, Some(base));
```
nicoburns: Ah, I see the confusion. These lines can be removed entirely. The call to `register_texture` is *instead* of the call to `override_image` (same for `unregister_texture`).

examples/wgpu_texture/src/demo_renderer.rs
```diff
@@ -172,7 +172,7 @@ impl ActiveDemoRenderer {
         // If "next texture" size doesn't match specified size then unregister and drop texture
         if let Some(next) = &self.next_texture {
             if next.texture.width() != width || next.texture.height() != height {
-                ctx.unregister_texture(next.handle);
+                ctx.unregister_texture(next.handle.clone());
                 self.next_texture = None;
             }
```
nicoburns: Probably not a big deal, but this could become:

```rust
if self.next_texture.is_some_and(|tex| tex.width() != width || tex.height() != height) {
      let handle = self.next_texture().take().unwrap().handle;
      ctx.unregister_texture(handle);
}
```

to avoid the clone.

nicoburns:
Vello `0.6.0` and Vello CPU `0.0.3` have now been released! I'm expecting a Parley release on Monday.

packages/anyrender_vello/src/wgpu_context.rs
rydb: > Fallback to CPU rendering is possible, but I wouldn't expect it to happen here. I'd expect it to happen if/when creation of a WGPU `Device` or `Adapter` fails during setup.

If thats the case, having the infrastructure for blitz to fail gracefully as some part of future error control flow to disable the UI rather then `panic!` is good for stability downstream. E.G: A bevy game that uses `dioxus-native` is in the middle of editing a large save file, GPU connection gets lost for some reason during the process, blitz forces a panic after a failed poll, and the save file gets corrupted.

At the end of the day though, its your call. I'll revert it to an .unwrap() if you'd rather it be that.

packages/anyrender_vello/src/custom_paint_source.rs
rydb: > Ah, I see the confusion. These lines can be removed entirely. The call to `register_texture` is _instead_ of the call to `override_image` (same for `unregister_texture`).

which lines?
```rust
    pub fn register_texture(&mut self, texture: Texture) -> TextureHandle {
        let dummy_image = self.renderer.register_texture(texture.clone());

        let base = TexelCopyTextureInfoBase {
            texture: texture,
            mip_level: 0,
            origin: wgpu::Origin3d::ZERO,
            aspect: wgpu::TextureAspect::All,
        };
        self.renderer.override_image(&dummy_image, Some(base));

        TextureHandle(dummy_image)
    }

    pub fn unregister_texture(&mut self, handle: TextureHandle) {
        self.renderer.override_image(&handle.0, None);
    }
```

nicoburns: Those functions should become:

```rust
    pub fn register_texture(&mut self, texture: Texture) -> TextureHandle {
        TextureHandle(self.renderer.register_texture(texture))
    }

    pub fn unregister_texture(&mut self, handle: TextureHandle) {
        self.renderer.unregister_texture(handle.0);
    }
```

packages/anyrender_vello/src/wgpu_context.rs
nicoburns: From what I can tell, Bevy itself is not nearly so careful with panic safety, so any game wanting to ensure that a save file is always saved would need to use `catch_unwind` anyway. But if you want this function to return an error then... fine I guess. But I do need you to put the panic back in at a higher level. Because the current code will fail silently which is not good.

rydb: > From what I can tell, Bevy itself is not nearly so careful with panic safety, so any game wanting to ensure that a save file is always saved would need to use `catch_unwind` anyway. But if you want this function to return an error then... fine I guess. But I do need you to put the panic back in at a higher level. Because the current code will fail silently which is not good.

the code currently panics in the only place that it appears, is this fine?
```rust
  if let Ok(recv_result) =
      block_on_wgpu(&self.device, receiver.receive()).inspect_err(|err| {
          panic!("channel inaccessible: {:#}", err);
      })
  {
      let _ = recv_result.unwrap();
  }
  ```

nicoburns (approved):
I think this is good enough now. I've pushed a few fixes to make CI happy.

PR #285, opened 2025-10-29
Use color attribute for text cursor color in text inputs

IsseW:
The more correct solution for this would be to use the `caret-color` style attribute, but I think just this is an improvement from always drawing a black text cursor.

From https://developer.mozilla.org/en-US/docs/Web/CSS/caret-color#description:
> The auto value sets the insertion caret to currentColor, which is the [color](https://developer.mozilla.org/en-US/docs/Web/CSS/color) of the text that is being added or deleted.

### DIFF
--- a/packages/blitz-paint/src/render.rs
+++ b/packages/blitz-paint/src/render.rs
@@ -452,12 +452,13 @@ impl ElementCx<'_> {
                     );
                 }
                 if let Some(cursor) = input_data.editor.cursor_geometry(1.5) {
-                    let color = self.style.clone_color();
+                    // TODO: Use the `caret-color` attribute here if present.
+                    let color = self.style.get_inherited_text().color;
+
                     scene.fill(
                         Fill::NonZero,
                         transform,
-                        // TODO: Use the `caret-color` attribute here if present.
-                        Color::new(*color.raw_components()),
+                        color.as_srgb_color(),
                         None,
                         &convert_rect(&cursor),
                     );


packages/blitz-paint/src/render.rs
```diff
@@ -452,10 +452,12 @@ impl ElementCx<'_> {
                     );
                 }
                 if let Some(cursor) = input_data.editor.cursor_geometry(1.5) {
+                    let color = self.style.clone_color();
                     scene.fill(
                         Fill::NonZero,
                         transform,
-                        Color::BLACK,
+                        // TODO: Use the `caret-color` attribute here if present.
+                        Color::new(*color.raw_components()),
```
nicoburns: I believe just passing `raw_components` here is wrong. That assumes that the color spaces match, but they won't necessarily. We have correct color conversion here: https://github.com/DioxusLabs/blitz/blob/main/packages/blitz-paint/src/color.rs You may also need to called `to_absolute_color()` first.

You may also want to use `style.get_inherited_text().color` rather than cloning the color as the colour can be a `calc()` node or similar which would be expensive to clone.

nicoburns:
If `caret-color` is important to you then let me know and we can look at enabling that property in Stylo

packages/blitz-paint/src/render.rs
IsseW: Ahh makes sense, thanks for the detailed response :smile:

nicoburns:
(support for the `caret-color` property landed in https://github.com/DioxusLabs/blitz/commit/4be7733ebd6e8227626db1dfd436788c7fa8c541)

PR #294, opened 2025-11-28
Fix README.md run commands for examples

TSPMP:
README.md seemed to contain commands for running the examples from back when the examples/ folder had a different organization.

This PR updates README.md with run commands that work on the current examples. It also contains some minor other fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Talk to us in: the [#native](https://discord.gg/AnNPqT95pu) channel in the [Diox
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


nicoburns (changes requested):
Thanks for the PR. We definitely need to update the example commands in the README. However I have some different suggestions.

1. I suggest we swap points (2) and (3)
2. The command for "browser ui" should now be `cargo run --release --package browser`. This runs the new UI which has a URL bar so passing a URL isn't required
3. `cargo run --release --package readme ./README.md` should now be considered an "example"
4. The todomvc example still exists. The new command is `cargo run --release --package todomvc`.
5. The other example I would recommend that we showcase is the `wgpu_texture` example which showcases integration with raw WGPU rendering.

nicoburns (approved):
Looks good

PR #302, opened 2025-12-12
feat: Window Manager APIs

JakkuSakura:
This PR introduces multi-window management and easy to use API and example


packages/dioxus-native/src/dioxus_application.rs
```diff
@@ -1,14 +1,65 @@
...
+
+impl<T: ?Sized> UnsafeBox<T> {
+    pub fn new(value: Box<T>) -> Self {
+        Self(value)
+    }
+
+    pub fn into_inner(self) -> Box<T> {
+        self.0
+    }
+}
```
nicoburns, line 91: Rather than this unsafe abstraction, could we have a "checked" Send/Sync box that uses https://doc.rust-lang.org/stable/std/thread/struct.ThreadId.html to check at runtime that box is created and consumed on the same thread?

packages/dioxus-native/Cargo.toml
```diff
@@ -94,8 +94,8 @@ manganis = { workspace = true, features = ["dioxus"], optional = true }
 winit = { workspace = true }
 keyboard-types = { workspace = true }
 
-# IO & Networking
-tokio = { workspace = true, features = ["rt-multi-thread"], optional = true }
+# Runtime + channels
+tokio = { workspace = true, features = ["rt-multi-thread", "sync"] }
```
nicoburns: Dioxus Desktop has `tokio` as an optional dependency and uses https://docs.rs/futures-channel for channels. Could we do the same in this PR?

examples/multi_window.rs
```diff
@@ -0,0 +1,83 @@
...
+            div {
+                button {
+                    onclick: move |_| {
+                        let vdom = VirtualDom::new(secondary_window);
+                        let title = format!("window#{}", counter());
+                        let attributes = WindowAttributes::default()
+                            .with_title(title)
+                            .with_inner_size(winit::dpi::LogicalSize::new(400.0, 300.0));
+                        let receiver = provider.create_document_window(vdom, attributes);
+                        let mut spawned_windows = spawned_windows.clone();
```
nicoburns: I wonder whether `create_document_window` should accept a "function that returns a `VirtualDom`" (similar to `launch`) rather than an already constructed `VirtualDom`? That might eliminate the thread-safety issues?

nicoburns (changes requested):
I have some comments about thread synchronisation. In addition to that, we're hoping to align the API with the one from Dioxus Desktop, so that users could use the same code with either Desktop or Native.

https://docs.rs/dioxus-desktop/latest/dioxus_desktop/fn.use_window.html

https://docs.rs/dioxus-desktop/latest/dioxus_desktop/struct.DesktopService.html#method.new_window

examples/multi_window.rs
JakkuSakura: Maybe not able to elimate thread-safety issues. It's because BlitzShellEvent::embedder_event still requires Send + Sync

PR #304, opened 2025-12-14
Add extra mouse event handling

rhaskia:
The following events now work with attributes:

- oncontextmenu (no actual context menu support)
- ondblclick
- onmouseover
- onmouseout
- onmouseleave
- onmouseenter
- onscroll
- onwheel

I did move the actual scroll handling code into handle_scroll, but it means redraw happens on every wheel event now, which is not ideal.

### DIFF
--- a/packages/blitz-shell/src/window.rs
+++ b/packages/blitz-shell/src/window.rs
@@ -401,7 +401,6 @@ impl<Rend: WindowRenderer> View<Rend> {
                 };
 
                 self.doc.handle_ui_event(UiEvent::Wheel(event));
-                self.request_redraw();
             }
 
             // File events


packages/blitz-dom/src/events/mouse.rs
```diff
@@ -275,4 +287,30 @@ pub(crate) fn handle_click<F: FnMut(DomEvent)>(
...
     doc.clear_focus();
+
+    // Assumed double click time to be less than 500ms, although may be system-dependant?
+    if doc.dbl_click_first_time.map(|t| t.elapsed() < Duration::from_millis(500)).unwrap_or(false)  {
+        dispatch_event(DomEvent::new(target, DomEventData::DoubleClick(event.clone())));
+        doc.dbl_click_first_time = None;
+    } else {
+        doc.dbl_click_first_time = Some(Instant::now());
+    }
 }
```
nicoburns, line 317: OS's do have a system setting for this. But I think assuming 500ms is fine for now, we can add that later.

What I think we will want is to generalise the "double click" tracking into a `click_count: u16` and `last_click_time`. Where we keep incrementing the count while clicks are still within the time limit (from `last_click_time`) and "double click" corresponds to click count 2.

Try clicking these examples many times in Chrome/Firefox:

- https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event#result
- https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#result

packages/blitz-dom/src/events/driver.rs
```diff
@@ -55,8 +55,20 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
...
+                        self.handle_dom_event(DomEvent::new(target, DomEventData::MouseLeave(event.clone())));
+                    }
+                }
                 hover_node_id = self.doc().hover_node_id;
+                if changed {
+                    if let Some(target) = hover_node_id {
+                        self.handle_dom_event(DomEvent::new(target, DomEventData::MouseOver(event.clone())));
+                        self.handle_dom_event(DomEvent::new(target, DomEventData::MouseEnter(event.clone())));
+                    }
+                }
```
nicoburns, line 120: This looks correct for `mouseover`/`mouseout`. But I think for `mouseenter`/`mouseleave` you actually need to diff the "node stack" against the previous "node stack", and individually send the event to each node that has been entered or left.

See: https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseenter_event#usage_notes

packages/blitz-shell/src/window.rs
```diff
@@ -383,20 +384,22 @@ impl<Rend: WindowRenderer> View<Rend> {
...
+                    button: MouseEventButton::Main, // Acts as an uninitialized value in this case
+                    buttons: self.buttons,
+                    mods: winit_modifiers_to_kbt_modifiers(self.keyboard_modifiers.state()),
                 };
 
-                if has_changed {
-                    self.request_redraw();
-                }
+                self.doc.handle_ui_event(UiEvent::Wheel(event));
+                self.request_redraw();
```
nicoburns: To fix the issue with every scroll event triggering a redraw, don't call `request_redraw` here. Instead have the event handler in the Document call `self.shell_provider.request_redraw()`.

See: the `MouseMove` case in `blitz-dom/src/events/mod.rs`

packages/blitz-dom/src/events/mouse.rs
nicoburns: Non-blocking, but we may want to consider generating `Instant::now()` in `blitz-shell` and have `blitz-dom`'s event handling functions take it as a parameter (/make it a field in the event struct). This would be really useful for testing things like click detection, and would also allow us to make use of event timestamps provided by the OS if/when those are available from Winit.

nicoburns (changes requested):

rhaskia:
Fixed those issues with this newest commit, although the Instant is still where it was initially

packages/blitz-dom/src/events/driver.rs
```diff
@@ -56,17 +56,49 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
                 let dom_x = event.x + viewport_scroll.x as f32 / zoom;
                 let dom_y = event.y + viewport_scroll.y as f32 / zoom;
                 let changed = self.doc_mut().set_hover_to(dom_x, dom_y);
+
+                let old_chain = hover_node_id.map(|id| self.doc().node_chain(id));
+                let new_chain = self.doc().hover_node_id.map(|id| self.doc().node_chain(id));
```
nicoburns: Try using `unwrap_or_default()` here to return an empty Vec in the case where the `hover_node_id` is `None`. Then the logic below doesn't need to deal with `Option`

packages/blitz-dom/src/events/driver.rs
```diff
@@ -55,8 +55,52 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
...
+
+                        // Send an mouseleave event to all old elements on the chain
+                        if let Some(ref old_chain) = old_chain {
+                            for i in chain_diff..old_chain.len() {
+                                self.handle_dom_event(DomEvent::new(old_chain[old_chain.len() - i - 1], DomEventData::MouseLeave(event.clone())));
+                            }
+                        }                         
+                    }
+                }
                 hover_node_id = self.doc().hover_node_id;
```
nicoburns: If you move up the  `hover_node_id = self.doc().hover_node_id;` line up, then everything from `let old_chain = ...` could be guarded by `if changed`

packages/blitz-dom/src/events/driver.rs
```diff
@@ -55,8 +55,52 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
...
+                        let old_len = old_chain.len();
+                        let new_len = new_chain.len();
+
+                        for i in 0..(old_len.min(new_len)) {
+                            if old_chain[old_len - i - 1] != new_chain[new_len - i - 1] {
+                                chain_diff = i;
+                            }
+                        }
+                    }
+                }
```
nicoburns: I don't think this will work in the case that the only difference is because one chain is longer than the other.

Once you've applied `unwrap_or_default` from above (so `old_chain` and `new_chain` are `Vec<T>`), I think the following would work (but should probably be tested):

```rust
let chain_diff = old_chain.iter().zip(&new_chain).position(|(old, new)| old != new).unwrap_or_else(|| old_len.min(new_len))
```

Aside: perhaps `chain_diff` should be called `first_different_index`?

packages/blitz-dom/src/events/driver.rs
```diff
@@ -55,8 +55,52 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
...
+
+                if changed {
+                    if let Some(target) = hover_node_id {
+                        self.handle_dom_event(DomEvent::new(target, DomEventData::MouseOut(event.clone())));
+
+                        // Send an mouseleave event to all old elements on the chain
+                        if let Some(ref old_chain) = old_chain {
+                            for i in chain_diff..old_chain.len() {
+                                self.handle_dom_event(DomEvent::new(old_chain[old_chain.len() - i - 1], DomEventData::MouseLeave(event.clone())));
+                            }
```
nicoburns: This needs to handle the `chain_diff > old_chain.len()` case. I'm not sure that it currently does.

I would recommend iterating over `old_chain.get(chain_diff..).unwrap_or(&[]).iter().rev()`

packages/blitz-dom/src/events/driver.rs
```diff
@@ -55,8 +55,52 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
...
+
+                if changed {
+                    if let Some(target) = hover_node_id {
+                        self.handle_dom_event(DomEvent::new(target, DomEventData::MouseOver(event.clone())));
+
+                        // Send an mouseenter event to all new elements on the chain
+                        if let Some(ref new_chain) = new_chain {
+                            for i in chain_diff..new_chain.len() {
+                                self.handle_dom_event(DomEvent::new(new_chain[new_chain.len() - i - 1], DomEventData::MouseEnter(event.clone())));
+                            }
```
nicoburns: This needs to handle the `chain_diff > new_chain.len()` case. I'm not sure that it currently does.

I would recommend iterating over `new_chain.get(chain_diff..).unwrap_or(&[]).iter().rev()`

packages/blitz-dom/src/events/driver.rs
rhaskia: Yes, that looks a lot cleaner. I was mostly doing it this way as node_chain works up from the deepest nested element but if I just reverse the chains before I use them that should work.

packages/blitz-dom/src/events/driver.rs
rhaskia: Never knew you could use partial ranges with .get. I'll add that in

packages/blitz-shell/src/window.rs
nicoburns: @rhaskia You seem to have the `request_redraw` in the wheel event handler, but not removed it from here?

packages/blitz-dom/src/document.rs
```diff
@@ -156,6 +151,12 @@ pub struct BaseDocument {
     pub(crate) active_node_id: Option<usize>,
     /// The node which recieved a mousedown event (if any)
     pub(crate) mousedown_node_id: Option<usize>,
+    /// The last time a click was made
+    pub(crate) last_click_time: Option<Instant>,
+    /// How many clicks have been made in quick succession
+    pub(crate) click_count: u16,
+    /// A list of nodes entered but not left
+    pub(crate) enter_stack: Vec<usize>,
```
nicoburns: Is `enter_stack` used? It doesn't look like it?

nicoburns (commented):
I've left comments on a couple of important-but-trivial nits. And clippy/rustfmt need to be fixed. Other than that this it looking pretty close now?

packages/blitz-dom/src/document.rs
rhaskia: My bad, I was going to use that for manually creating a stack of what had been entered, but I switched to node_chain and forgot about it.

PR #315, opened 2025-12-23
Text selection for non-input elements

### LINKED ISSUE #227: Text selection
Blitz should support text selection (outside of text inputs where it is already supported). This would allow users to copy text out of apps/pages rendered with Blitz. This initial version of the functionality should probably focus on the "easy" case of text selection within a single inline formatting context (which will be backed by a single Parley `Layout`).

Himmelschmidt:
Initial implementation of text selection outside of text inputs (fixes #227).

Adds click-and-drag selection and Ctrl+C/Cmd+C to copy. Works across multiple inline formatting contexts and handles anonymous blocks.

Still a bit rough around the edges, let me know if this is the right direction or too big of an initial go at it.

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -152,6 +152,8 @@ pub struct BaseDocument {
 
     /// The node which is currently hovered (if any)
     pub(crate) hover_node_id: Option<usize>,
+    /// Whether the node which is currently hovered is a text node/span
+    pub(crate) hover_node_is_text: bool,
     /// The node which is currently focussed (if any)
     pub(crate) focus_node_id: Option<usize>,
     /// The node which is currently active (if any)
@@ -314,6 +316,7 @@ impl BaseDocument {
             layout_ctx: parley::LayoutContext::new(),
 
             hover_node_id: None,
+            hover_node_is_text: false,
             focus_node_id: None,
             active_node_id: None,
             mousedown_node_id: None,
@@ -1021,6 +1024,7 @@ impl BaseDocument {
     pub fn set_hover_to(&mut self, x: f32, y: f32) -> bool {
         let hit = self.hit(x, y);
         let hover_node_id = hit.map(|hit| hit.node_id);
+        let new_is_text = hit.map(|hit| hit.is_text).unwrap_or(false);
 
         // Return early if the new node is the same as the already-hovered node
         if hover_node_id == self.hover_node_id {
@@ -1042,6 +1046,7 @@ impl BaseDocument {
         }
 
         self.hover_node_id = hover_node_id;
+        self.hover_node_is_text = new_is_text;
 
         // Update the cursor
         let cursor = self.get_cursor().unwrap_or_default();
@@ -1120,7 +1125,6 @@ impl BaseDocument {
     }
 
     pub fn get_cursor(&self) -> Option<CursorIcon> {
-        // todo: cache this on the node itself
         let node = &self.nodes[self.get_hover_node_id()?];
 
         if let Some(subdoc) = node.element_data().and_then(|el| el.sub_doc_data()) {
@@ -1135,11 +1139,10 @@ impl BaseDocument {
             return Some(keyword);
         }
 
-        // Return text cursor for text nodes and text inputs
-        if node.is_text_node()
-            || node
-                .element_data()
-                .is_some_and(|e| e.text_input_data().is_some())
+        // Return text cursor for text inputs
+        if node
+            .element_data()
+            .is_some_and(|e| e.text_input_data().is_some())
         {
             return Some(CursorIcon::Text);
         }
@@ -1152,6 +1155,11 @@ impl BaseDocument {
             }
 
             maybe_node = node.layout_parent.get().map(|node_id| node.with(node_id));
+        }
+
+        // Return text cursor for text nodes
+        if self.hover_node_is_text {
+            return Some(CursorIcon::Text);
         }
 
         // Else fallback to default cursor
--- a/packages/blitz-dom/src/node/node.rs
+++ b/packages/blitz-dom/src/node/node.rs
@@ -909,7 +909,12 @@ impl Node {
                 {
                     let style_index = cluster.glyphs().next()?.style_index();
                     let node_id = layout.styles()[style_index].brush.id;
-                    return Some(HitResult { node_id, x, y });
+                    return Some(HitResult {
+                        node_id,
+                        x,
+                        y,
+                        is_text: true,
+                    });
                 }
             }
         }
@@ -920,6 +925,7 @@ impl Node {
                 node_id: self.id,
                 x,
                 y,
+                is_text: false,
             });
         }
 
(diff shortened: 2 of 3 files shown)


packages/blitz-paint/src/text.rs
```diff
@@ -87,3 +87,31 @@ pub(crate) fn stroke_text<'a>(
...
+    let selection_color = color::palette::css::STEEL_BLUE;
+    let scale = layout.scale() as f64;
+
+    selection.geometry_with(layout, |rect, _line_idx| {
+        // Convert from Parley's BoundingBox to kurbo::Rect, accounting for scale
+        let rect = kurbo::Rect::new(
+            rect.x0 as f64 / scale,
+            rect.y0 as f64 / scale,
+            rect.x1 as f64 / scale,
+            rect.y1 as f64 / scale,
```
nicoburns: These `/ scale` are incorrect. Parley dimensions are in physical pixels, and painting code draws in physical pixels. So you can just remove the conversion entirely.

packages/blitz-paint/src/text.rs
```diff
@@ -87,3 +87,31 @@ pub(crate) fn stroke_text<'a>(
...
+    layout: &Layout<TextBrush>,
+    transform: Affine,
+    selection_start: usize,
+    selection_end: usize,
+) {
+    let anchor = Cursor::from_byte_index(layout, selection_start, Affinity::Downstream);
+    let focus = Cursor::from_byte_index(layout, selection_end, Affinity::Downstream);
+    let selection = Selection::new(anchor, focus);
+
+    let selection_color = color::palette::css::STEEL_BLUE;
```
nicoburns: Can we change this color to `Color::from_rgb8(180, 213, 255)`? This matches the selection color in Safari and Firefox use (and is very close to the color Chrome uses) on macOS (can we also switch this color for text input selections).

I realise that the `STEEL_BLUE` color was already in use, but it looks bad when selecting page text.

packages/blitz-dom/src/events/mouse.rs
```diff
@@ -84,84 +90,121 @@ pub(crate) fn handle_mousemove<F: FnMut(DomEvent)>(
...
-            let layout = text_input_data.editor.try_layout().unwrap();
-            let content_box_height = node.final_layout.content_box_height();
-            let input_height = layout.height() / layout.scale();
-            let y_offset = ((content_box_height - input_height) / 2.0).max(0.0);
-
-            content_box_offset.y += y_offset;
+    let click_target = {
+        let node = &doc.nodes[actual_target];
+        match node.data.downcast_element() {
+            Some(el) if el.attr(local_name!("disabled")).is_some() => ClickTarget::Disabled,
```
nicoburns: Nit: `el.has_attr(local_name!("disabled"))` would be slightly nicer.

packages/blitz-dom/src/layout/construct.rs
```diff
@@ -287,6 +287,10 @@ pub(crate) fn collect_layout_children(
                 doc.nodes[container_node_id]
                     .flags
                     .insert(NodeFlags::IS_INLINE_ROOT);
+
+                // Set layout_parent for all inline children so they can find their inline root
+                set_inline_children_layout_parent(doc, container_node_id, container_node_id);
+
```
nicoburns: This function should perhaps be combined with `find_inline_layout_embedded_boxes`. As it is doing the same traversal.

packages/blitz-dom/src/selection.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct SelectionEndpoint {
+    /// The inline root node containing this endpoint
+    pub node: Option<usize>,
+    /// Byte offset within the inline root's text
+    pub offset: usize,
+    /// Parent element ID (stable reference for anonymous blocks)
+    pub parent: Option<usize>,
+    /// Index among anonymous siblings (for anonymous blocks only)
+    pub sibling_index: Option<usize>,
+}
```
nicoburns, line 22: We could probably combine `node` and `parent` into one `usize` field, and then interpret it as "node" if `sibling_index` is `None` and as "parent" if `sibling_index` is `Some`?

nicoburns: Also, I believe that it may be possible for there to be an anonymous node with other anonymous nodes as parents. So it may be that `sibling_index` needs to become `node_path: SmallVec<[usize; 1]>` or similar.

(this could probably left as a follow-up)

packages/blitz-dom/src/traversal.rs
```diff
@@ -235,4 +237,223 @@ impl BaseDocument {
...
+    }
+
+    /// Collect all inline root nodes between start_node and end_node in document order.
+    /// Both start and end are assumed to be inline roots.
+    /// Returns the nodes in document order (from first to last).
+    pub fn collect_inline_roots_in_range(&self, start_node: usize, end_node: usize) -> Vec<usize> {
+        let start = &self.nodes[start_node];
+        let end = &self.nodes[end_node];
+        let start_is_anon = start.is_anonymous();
+        let end_is_anon = end.is_anonymous();
```
nicoburns: For this function, would it make sense to just resolve the current node id of the anonymous node? Then the following code wouldn't have to treat anonymous nodes specially?

packages/blitz-dom/src/document.rs
```diff
@@ -1320,6 +1325,287 @@ impl BaseDocument {
...
+
+        let mut anon_index = 0;
+        for &child_id in children.iter() {
+            if self.get_node(child_id).is_some_and(|n| n.is_anonymous()) {
+                if anon_index == target_index {
+                    return Some(child_id);
+                }
+                anon_index += 1;
+            }
+        }
```
nicoburns: You could use `.filter(...).position(...)` for this if you wanted to use an iterator-based approach.

nicoburns (changes requested):
I have requested some changes, but this generally looks excellent. I think we can probably get this merged soon!

Of the individual comments, the issue with the scale factor is probably the only thing that's blocking.

There is also one other issue that I'd like to see fixed: currently when you select text and your selection ends on an element with a click action (e.g. a link) then that click action will be triggered (so if you select text ending in a link then you can end up navigating to another page!).

I think we should make it so that a given mouse action either causes a text selection XOR a click. I would suggest:

- Recording the x/y position on `mousedown`
- Only processing `mousemove` events as a selection if the cursor position is at least 2px in one direction away from the mousedown position. If that does occur then set a global "selecting" state on the document.
- Only generate a click event on `mouseup` if not in the "selecting" state.

We're already doing a similar 2px box for detecting double clicks, and it may be possible to combine these

packages/blitz-dom/src/layout/construct.rs
Himmelschmidt: Missed that, values it was setting were being overwritten by `find_inline_layout_embedded_boxes`, I'll remove it

Himmelschmidt:
I've pushed some updates according to your feedback, also made some improvements to handling anonymous nodes. I did see an edge case with text selection when a block element is nested inside `display: contents`. Selecting across the block causes incorrect highlighting (text after the block highlights prematurely, then unhighlights).

I think the root cause is that `display: contents` promotes children to the parent, causing inline text before and after an intervening block to be combined into a single inline root. This makes it impossible for selection logic to determine which portions to highlight.

This probably needs a deeper fix in layout/box generation to split inline text around blocks in `display: contents` contexts, I'll leave that for another time.

nicoburns (approved):
Sounds like there are still some issues in edge cases, but I think this an excellent first cut :)

PR #319, opened 2025-12-25
feat(blitz-dom): improve disabled attribute support

cschierig:
Adds improved support for the `disabled` attribute on form controls.

The following parts of the HTML spec are not yet supported:

- Inherit disabled status (depends on `select` element support)
- Disable custom form controls (I'm not sure if those are supported by Blitz)


nicoburns (approved):
Looks good - thanks!

PR #343, opened 2026-02-06
fix: check for visibility changes for damage

### LINKED ISSUE #335: Elements that show on `:hover` don't always show when incremental layout is enabled
### To reproduce

- Run `just todomvc --features incremental`. The X to delete a todo does not show.
- OR Run `just incr https://old.reddit.com/r/rust`. The warning when hovering a "downvote" button does not show.

Both of these functionalities work with incremental layout disabled.

kokoISnoTarget:
This fixes #335, by adding restyle damage on visibility changes.


kokoISnoTarget:
It seems like it at least needs `CONSTRUCT_BOX` as RestyleDamage for this to work.

nicoburns:
Hmm... this still isn't quite right. Without incremental mode the box animates in and out, and the text is a clear bright white. With incremental mode + this PR there's no animation and the text is very faint.

PR #361, opened 2026-03-02
Feature/errors in browser UI

mogambro:
Won't win any awards...

image

### DIFF
--- a/apps/browser/src/main.rs
+++ b/apps/browser/src/main.rs
@@ -420,12 +420,7 @@ impl DocumentLoader {
                 Err(err) => {
                     println!("Error loading document {:?}", err);
 
-                    let error_msg = format!("{err:?}")
-                        .replace('&', "&amp;")
-                        .replace('<', "&lt;")
-                        .replace('>', "&gt;");
-                    let error_html =
-                        include_str!("../assets/error.html").replace("{error}", &error_msg);
+                    let error_msg = format!("{err:?}");
 
                     let config = DocumentConfig {
                         viewport: None,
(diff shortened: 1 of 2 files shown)


apps/browser/src/main.rs
```diff
@@ -419,6 +419,27 @@ impl DocumentLoader {
                 }
                 Err(err) => {
                     println!("Error loading document {:?}", err);
+
+                    let error_msg = format!("{err:?}")
```
mogambro: We could use a sanitizer library here instead... we probably should.

Something like https://lib.rs/crates/hescape ?

Dunno, it's very new and I have no prior experience with such libraries in Rust

apps/browser/src/main.rs
```diff
@@ -419,6 +419,27 @@ impl DocumentLoader {
                 }
                 Err(err) => {
                     println!("Error loading document {:?}", err);
```
mogambro, line 421: At some point we should probably switch all println! here to tracing equivalents... or is there a non-historical reason for the println!'s?

apps/browser/src/main.rs
nicoburns: You could load the HTML without the error string, and just put a div with an ID. Then you get_element_by_id and set_text to set the error message.

Alternatively if you created the error page as a Dioxus component rather than an HTML page then we could could take advantage of the escaping support in the Dioxus layer.

mogambro: What do you prefer? I am mostly interested in non-dioxus/bare-bones browser at the moment.

A tiny ladybird alternative, so to speak :)

nicoburns: I think I'd prefer the Dioxus component. My reasoning being that this would make it easy to add interactive functionality to the error page if desired (e.g. a "Try Again" button).

Note that the Browser App UI is written in Dioxus Native. This isn't something we're hugely committed (the Browser App is <1000 LoC), but it is quite nice to be able to render the UI using the same underlying renderer (Blitz) as is used to render the web page content.

I imagine that a browser UI using an alternative GUI toolkit would also manage it's own error pages.

apps/browser/src/main.rs
nicoburns: No reason for `println`

Issue #363, opened 2026-03-03
Browser UI features


nicoburns:
These are features that could be implemented in the Blitz browser UI (`/apps/browser`).

 ### Browser UI Backlog

- [ ] Menu items:
    - [ ] Toggle incremental layout
    - [x] Toggle element inspector
    - [ ] Toggle element outlines
    - [ ] Capture AnyRender scene
    - [x] Capture screenshot
    - [x] View source
    - [ ] Clear HTTP cache
- [ ] Built-in pages
    - [ ] Custom start page
    - [ ] Generic error page?
    - [x] 404 page
- [ ] Status bar
- [ ] Tabs (multiple embedded documents)
    - [ ] https://github.com/adamschwartz/chrome-tabs
    - [ ] Draggable: https://gist.github.com/alexreardon/9ef479804a7519f713fe2274e076f1f3
- [ ] Reader mode
- [ ] History + URL suggest (nucleo crate)
- [ ] Find in page
- [ ] FPS counter
- [ ] Fragment links (scroll in page)
- [ ] Bug: browser redraws after every mousemove/scroll event
- [ ] Bug with scale factor: text on wikipedia gets too large when resizing the window
- [ ] Ad blocking using https://github.com/brave/adblock-rust

tonybierman:
# Browser UI Features (#363) — Proposed Implementation Order

## Principles

1. **Bugs before features** — performance/correctness issues compound and infect anything built on top.
2. **Foundations before dependents** — a status bar and built-in pages mechanism host later features.
3. **Cheap, isolated wins next** — ship value while the spec is still warm.
4. **Architecturally invasive items last** — tabs touch the document model; doing them last means each prior feature is built once on a single doc, then mechanically lifted into the tab wrapper.

## Proposed Order

| # | Item | Why here |
|---|------|----------|
| 1 | **Bug: redraws on every mousemove/scroll** | Performance bug. Every later feature will be measured against current frame timing — fix first or you debug it forever. |
| 2 | **Bug: scale factor text resize on Wikipedia** | Correctness bug, visible. Fix before adding more UI that interacts with scaling. |
| 3 | **Status bar** | Chrome foundation. Find-in-page, loading state, URL hover, etc. all want to render into it. Build the host once. |
| 4 | **Generic error page** | Tiny. Reuses the same built-in-page plumbing as the existing 404 page. Good warmup that exercises that path. |
[...]

PR #364, opened 2026-03-03
Implement the view-source menu-item from #363

mogambro:
This allows to, well, view the source code of the current document.

Trade-Off: We store the entire source as a Signal<String> ... an alternative would be to refetch it.

Screenshot from 2026-03-03 20-19-30
Screenshot from 2026-03-03 20-19-50

### DIFF
--- a/apps/browser/src/main.rs
+++ b/apps/browser/src/main.rs
@@ -107,13 +107,7 @@ fn app() -> Element {
         let current_url = history.current_url().read().url.to_string();
         *url_input_value.write() = format!("view-source://{current_url}");
 
-        let escaped = source
-            .replace('&', "&amp;")
-            .replace('<', "&lt;")
-            .replace('>', "&gt;");
-        let view_source_html = format!(
-            "<html><body><pre style=\"margin:8px;white-space:pre-wrap;word-wrap:break-word;\"><code>{escaped}</code></pre></body></html>"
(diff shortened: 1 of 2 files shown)


apps/browser/src/main.rs
```diff
@@ -412,7 +471,7 @@ impl DocumentLoader {
                         str::from_utf8(&bytes).unwrap()
                     };
 
-                    // println!("{}", html);
+                    *html_source.write_unchecked() = html.to_string();
```
mogambro, line 473: no highlighting for now - would create another ticket for that.

apps/browser/src/main.rs
```diff
@@ -91,8 +94,42 @@ fn app() -> Element {
...
+        let current_url = history.current_url().read().url.to_string();
+        *url_input_value.write() = format!("view-source://{current_url}");
+
+        let escaped = source
+            .replace('&', "&amp;")
+            .replace('<', "&lt;")
+            .replace('>', "&gt;");
+        let view_source_html = format!(
+            "<html><body><pre style=\"margin:8px;white-space:pre-wrap;word-wrap:break-word;\"><code>{escaped}</code></pre></body></html>"
+        );
```
nicoburns: This has the same escaping issue as the error page did. Need to use `set_text`.

PR #365, opened 2026-03-03
Feature/capture screenshot

mogambro:
Screenshot from 2026-03-03 20-43-01
Screenshot from 2026-03-03 20-43-12

### DIFF
--- a/apps/browser/Cargo.toml
+++ b/apps/browser/Cargo.toml
@@ -11,7 +11,7 @@ path = "src/main.rs"
 
 [features]
 default = ["vello", "floats", "incremental", "cookies", "cache", "screenshot"]
-screenshot = ["dep:anyrender", "dep:anyrender_vello_cpu", "dep:blitz-paint", "dep:png", "dep:peniko"]
+screenshot = ["dep:anyrender", "dep:anyrender_vello_cpu", "dep:blitz-paint", "dep:png", "dep:peniko", "dep:rfd"]
 vello = ["dioxus-native/vello"]
 hybrid = ["dioxus-native/vello-hybrid"]
 skia = ["dioxus-native/skia"]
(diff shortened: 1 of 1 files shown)


apps/browser/src/main.rs
```diff
@@ -91,8 +100,102 @@ fn app() -> Element {
...
+                    render_height,
+                );
+
+                // Save to file
+                let timestamp = std::time::SystemTime::now()
+                    .duration_since(std::time::UNIX_EPOCH)
+                    .unwrap()
+                    .as_secs();
+                let filename = format!("blitz-screenshot-{timestamp}.png");
+                if let Ok(file) = std::fs::File::create(&filename) {
```
mogambro: I am not sure where we should store screenshots... I am on Linux, so it could be /home/mogambro/pictures/screenshots or /home/mogambro/downloads or anything, really...

Not sure whether there is a XDG standard for that.

Not to speak about MacOS or that other OS from Microslop :)

apps/browser/Cargo.toml
```diff
@@ -49,6 +49,11 @@ tracing-subscriber = { workspace = true, optional = true }
 webbrowser = { workspace = true }
 winit = { workspace = true }
 image = { workspace = true }
+anyrender = { workspace = true }
+anyrender_vello_cpu = { workspace = true }
+blitz-paint = { workspace = true }
+png = { workspace = true }
+peniko = { workspace = true }
```
nicoburns, line 52: Can we feature-flag the screenshot functionality?

apps/browser/src/main.rs
nicoburns: Perhaps we could use https://docs.rs/rfd to let the user select a location?

apps/browser/src/main.rs
```diff
@@ -128,6 +134,58 @@ fn app() -> Element {
...
+        menu_open.set(false);
+        if let Some(handle) = webview_node_handle() {
+            let node_id = handle.node_id();
+            let mut doc = handle.doc_mut();
+            if let Some(sub_doc) = doc
+                .get_node_mut(node_id)
+                .and_then(|node| node.element_data_mut())
+                .and_then(|el| el.sub_doc_data_mut())
+            {
+                let sub_doc = sub_doc.inner();
```
nicoburns: Perhaps this bit that actually creates the screenshot ought to be it's own function?

nicoburns (approved):
For some reason the sync file dialog was panicking for me on macOS. So I've changed this to use the async file dialog. Otherwise this looks good :)

PR #366, opened 2026-03-06
Draw borders with the same color in single bezpath.

### LINKED ISSUE #359: Borders Have Artifacts
```html

<!DOCTYPE html>

<html>

 <head>

  <style type="text/css">
  div {border: green solid 8px;}
  </style>

 </head>

 <body>

  <div></div>

 </body>
</html>
```

Blitz:

Image

Chrome:

Image

AustinMReppert:
This is a small fix to ensure adjacent borders with the same color are drawn in a single bezpath to avoid artifacts.

Gemini was used to assist with the code.

Fixes #359 

prior:

Of tests run:
 201 tests CRASHED (0.77% of run; 0.31% of found)
12020 tests PASSED (45.93% of run; 18.48% of found)
13950 tests FAILED (53.30% of run; 21.45% of found)

with border change:

Of tests run:
 201 tests CRASHED (0.77% of run; 0.31% of found)
12130 tests PASSED (46.35% of run; 18.65% of found)
13840 tests FAILED (52.88% of run; 21.28% of found)

Prior:
Screenshot from 2026-03-05 23-42-42

With border Fix:

Screenshot from 2026-03-05 23-43-34

### DIFF
--- a/packages/blitz-paint/src/render.rs
+++ b/packages/blitz-paint/src/render.rs
@@ -810,71 +810,65 @@ impl ElementCx<'_> {
         let border = style.get_border();
         let current_color = style.clone_color();
 
-        let mut border_groups: [Option<(Color, BezPath)>; 4] = [None, None, None, None];
-        let mut border_group_count = 0;
+        let mut borders: [(Color, Option<BezPath>); 4] = [
+            (Color::TRANSPARENT, None),
+            (Color::TRANSPARENT, None),
+            (Color::TRANSPARENT, None),
+            (Color::TRANSPARENT, None),
+        ];
+        let mut count = 0;
 
-        for edge in [Edge::Top, Edge::Right, Edge::Bottom, Edge::Left] {
+        for &edge in &[Edge::Top, Edge::Right, Edge::Bottom, Edge::Left] {
             let color = match edge {
-                Edge::Top => border
-                    .border_top_color
-                    .resolve_to_absolute(&current_color)
-                    .as_srgb_color(),
-                Edge::Right => border
-                    .border_right_color
-                    .resolve_to_absolute(&current_color)
-                    .as_srgb_color(),
-                Edge::Bottom => border
-                    .border_bottom_color
-                    .resolve_to_absolute(&current_color)
-                    .as_srgb_color(),
-                Edge::Left => border
-                    .border_left_color
-                    .resolve_to_absolute(&current_color)
-                    .as_srgb_color(),
-            };
-
-            if color.components[3] == 0.0 {
-                continue;
+                Edge::Top => &border.border_top_color,
+                Edge::Right => &border.border_right_color,
+                Edge::Bottom => &border.border_bottom_color,
+                Edge::Left => &border.border_left_color,
             }
+            .resolve_to_absolute(&current_color)
+            .as_srgb_color();
 
-            let border_path = self.frame.border_edge_shape(edge);
-
-            // Check if we can merge with the current active group.
-            if border_group_count > 0 {
-                if let Some((last_color, last_path)) = &mut border_groups[border_group_count - 1] {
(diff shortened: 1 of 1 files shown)


packages/blitz-paint/src/render.rs
```diff
@@ -806,8 +806,69 @@ impl ElementCx<'_> {
...
+                scene.fill(Fill::NonZero, self.transform, color, None, &border_path);
+            } else {
+                scene.fill(
+                    Fill::NonZero,
+                    self.transform,
+                    color,
+                    None,
+                    borders[start_border_index].1.as_ref().unwrap(),
+                );
+            }
```
nicoburns, line 870: This can be simplified:

```rust
let mut next_border_index = start_border_index + 1;
let mut border_path = borders[start_border_index].1.take().unwrap();
while next_border_index < count && borders[next_border_index].0 == color {
    border_path.extend(&borders[next_border_index].1.take().unwrap());
    next_border_index += 1;
}
scene.fill(Fill::NonZero, self.transform, color, None, &border_path);
```

The key insight here being that the `has_multiple_edges` "if" is using the same condition as the `while` loop in the first branch of the if. So we simplify to only using a `while` loop that just does zero iterations in the "else" case.

packages/blitz-paint/src/render.rs
```diff
@@ -806,8 +806,69 @@ impl ElementCx<'_> {
...
+        let mut start_border_index = 0;
+        while start_border_index < count {
+            let color = borders[start_border_index].0;
+            let mut next_border_index = start_border_index + 1;
+            let has_multiple_edges =
+                next_border_index < count && borders[next_border_index].0 == color;
+            if has_multiple_edges {
+                let mut border_path = borders[start_border_index].1.take().unwrap();
+                while next_border_index < count && borders[next_border_index].0 == color {
+                    border_path.extend(&borders[next_border_index].1.take().unwrap());
```
nicoburns, line 858: The docs on `BezPath::extend` (https://docs.rs/kurbo/latest/kurbo/struct.BezPath.html#method.extend) say:

> Note that if you’re attempting to make a continuous path, you will generally want to ensure that the iterator does not contain any [MoveTo](https://docs.rs/kurbo/latest/kurbo/enum.PathEl.html#variant.MoveTo) or [ClosePath](https://docs.rs/kurbo/latest/kurbo/enum.PathEl.html#variant.ClosePath) elements. Note especially that many (open) [shapes](https://docs.rs/kurbo/latest/kurbo/trait.Shape.html) will start with a MoveTo if you use their [path_elements](https://docs.rs/kurbo/latest/kurbo/trait.Shape.html#tymethod.path_elements) function. Some shapes have alternatives for this use case, such as [Arc::append_iter](https://docs.rs/kurbo/latest/kurbo/struct.Arc.html#method.append_iter).

Which I think we are doing here?

I guess if it works this is fine. But it would be nice to understand if/why this is ok.

packages/blitz-paint/src/render.rs
```diff
@@ -806,8 +806,69 @@ impl ElementCx<'_> {
...
     /// Draw all borders for a node
     fn draw_border(&self, scene: &mut impl PaintScene) {
-        for edge in [Edge::Top, Edge::Right, Edge::Bottom, Edge::Left] {
-            self.draw_border_edge(scene, edge);
+        let style = &*self.style;
+        let border = style.get_border();
+        let current_color = style.clone_color();
+
+        let mut borders: [(Color, Option<BezPath>); 4] = [
+            (Color::TRANSPARENT, None),
```
nicoburns, line 814: Nit: It might be nice for readability to use a struct for this? So we have `.color` and `.path` rather than `.0` and `.1`...

nicoburns:
> Gemini was used to assist with the code.

I appreciate the disclosure!

PR #367, opened 2026-03-06
Add word-spacing.

AustinMReppert:
When running with the border fix this change should ensure this test passes.

wpt/css/CSS2/borders/border-003-ref.xht

Previous:
Screenshot from 2026-03-06 00-05-46

With Change:
Screenshot from 2026-03-06 00-04-45

### DIFF
--- a/packages/blitz-dom/src/stylo_to_parley.rs
+++ b/packages/blitz-dom/src/stylo_to_parley.rs
@@ -125,6 +125,11 @@ pub(crate) fn style(
         .resolve(Length::new(font_size))
         .px();
 
+    let word_spacing = itext_styles
+        .word_spacing
+        .resolve(Length::new(font_size))
+        .px();
+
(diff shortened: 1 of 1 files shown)


packages/blitz-dom/src/stylo_to_parley.rs
```diff
@@ -125,6 +125,11 @@ pub(crate) fn style(
         .resolve(Length::new(font_size))
         .px();
 
+    let word_spacing = itext_styles
+        .word_spacing
+        .resolve(Length::new(font_size))
```
nicoburns, line 130: I don't think `font_size` is the right thing to resolve these against. I believe it should be "the width of the space character". But seeing as we're already doing this incorrectly for `letter-spacing` I'll allow it for now.

nicoburns:
Thanks!

Issue #128, opened 2024-09-06
PDF renderer


nicoburns:
It would be a cool test of the pluggable renderers to create a renderer that renders to PDF.

We should consider using the https://github.com/LaurenzV/krilla crate for this.

mitsuru:
Hi @nicoburns — I've been building exactly this over at https://github.com/mitsuru/fulgur: an HTML/CSS → PDF engine on top of Blitz + Krilla.

It covers the full pipeline:

- HTML → Blitz (parse/style/layout) → Pageable tree → Krilla
- Page splitting with widow/orphan handling
- CSS Generated Content for Paged Media (`@page`, margin boxes, running elements, counters, `string-set`)
- Offline-first asset bundling (no network during render)

One thing that's come up while implementing print styling: `blitz-dom` currently hardcodes `MediaType::screen()` in `make_device` (`document.rs:180`), so `@media print { ... }` rules never apply. For PDF use cases this matters a lot. I'd be happy to send a small PR — e.g. adding a `set_media_type(MediaType)` method on `BaseDocument`, or a `media_type` field on `DocumentConfig` — if that direction aligns with where you want the API to go.

Thanks for Blitz — it's been a great foundation to build on.

nicoburns:
@mitsuru Very cool!

I'd definitely accept a PR for `set_media_type()` or similar. And if there's anything else you need then please feel free to submit issues and/or PRs.

It's not a small task, but eventually I'd like to get fragmented layout and paged media support into Blitz itself (and Taffy/Parley).

PR #390, opened 2026-04-14
Add BaseDocument::set_media_type for paged/print use cases

mitsuru:
Refs #128.

As discussed in https://github.com/DioxusLabs/blitz/issues/128#issuecomment-4246086994, this adds a way to configure the CSS media type on `BaseDocument` so that `@media print { ... }` rules can be evaluated. Previously `make_device` hardcoded `MediaType::screen()`, which blocks any paged / print-styled rendering (e.g. PDF) on top of Blitz.

## Changes

- `DocumentConfig::media_type: Option<MediaType>` — defaults to `MediaType::screen` when `None`.
- `BaseDocument` now stores the resolved `MediaType` and threads it through `make_device` (initial build, `set_viewport`, `ViewportMut` drop). The internal `make_device` takes `MediaType` by value — every call site already has a concrete value, so there's no `Option` wrapping / hidden default.
- New `BaseDocument::media_type()` (getter) and `BaseDocument::set_media_type(MediaType)` — the setter rebuilds the stylist device and force-dirties origins so rules re-evaluate on the next restyle. Early-returns if the media type is unchanged.
- `style::media_queries::MediaType` is re-exported from `blitz_dom` so downstream crates don't need a direct `stylo` dependency just to switch media types.
- Updated the three `DocumentConfig` struct literals in `apps/browser` to include `media_type: None`.

## Tests

Added unit tests in `packages/blitz-dom/src/mutator.rs`:

- `media_type_defaults_to_screen` — default `DocumentConfig` → `screen`.
- `media_type_honors_config` — `DocumentConfig { media_type: Some(MediaType::print()), .. }` → `print`.
- `set_media_type_updates_stylist_device` — after `set_media_type(print)`, both `document.media_type()` and `document.stylist_device().media_type()` return `print`.

Verified with `cargo check --workspace` and `cargo test -p blitz-dom --lib` (16 passed).

## Usage

```rust
use blitz_dom::{BaseDocument, DocumentConfig, MediaType};

// At construction
let doc = BaseDocument::new(DocumentConfig {
    media_type: Some(MediaType::print()),
    ..Default::default()
});

// Or switch at runtime
doc.set_media_type(MediaType::print());
```

Validated end-to-end on top of [fulgur](https://github.com/mitsuru/fulgur) (HTML/CSS → PDF via Blitz + Krilla): `@media print` rules now apply correctly in rendered PDFs.

### DIFF
--- a/apps/browser/src/main.rs
+++ b/apps/browser/src/main.rs
@@ -122,6 +122,7 @@ fn app() -> Element {
             shell_provider: None,
             html_parser_provider: Some(Arc::new(HtmlProvider)),
             font_ctx: Some(loader.font_ctx.clone()),
+            media_type: None,
         };
         let mut document = HtmlDocument::from_html(view_source_html, config).into_inner();
         if let Some(parent_id) = document.get_element_by_id("source") {
@@ -544,6 +545,7 @@ impl DocumentLoader {
                         shell_provider: Some(consume_context::<Arc<dyn ShellProvider>>()),
                         html_parser_provider: Some(Arc::new(HtmlProvider)),
                         font_ctx: Some(font_ctx),
+                        media_type: None,
                     };
 
                     let html = if bytes.is_empty() {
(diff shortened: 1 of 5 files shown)


nicoburns (approved):
Not sure if the tests are adding much value and may delete them at some point. But they're fairly short so 🤷

PR #396, opened 2026-04-25
Implement CSS counter-style symbols() support

lainon1:
## Summary
- Implements CSS `symbols()` function for list-style-type custom counter styles in list markers
- Previously the code fell back to • for all custom counter-style definitions
- Now properly cycles through defined symbols based on list item index per CSS spec

## Changes
- Modified `packages/blitz-dom/src/layout/list.rs` to handle `CounterStyle::Symbols` variant
- Handles both String and Ident symbol types from the stylo crate
- Empty symbols list falls back to • (preserves existing behavior)

## Related
- Fixes TODO comment in layout/list.rs:137


nicoburns (approved):
Looks good - thanks!

PR #397, opened 2026-04-25
perf: optimize box-shadow clipping

lainon1:
### 1. Box-shadow clip optimization (`box_shadow.rs`)

Previously, a clip/layer was always created when an element had an outset box-shadow, even if the element was fully opaque (`opacity == 1.0`). This is unnecessary because the shadow can only show through when there's transparency.

**Fix:** Only create the clip layer when `opacity < 1.0`, avoiding layer creation for the common case of fully opaque elements.

## Testing

- `cargo check -p blitz-paint` passes
- `cargo check -p wpt` passes
- `cargo test -p blitz-paint` passes

### DIFF
--- a/packages/blitz-paint/src/render/box_shadow.rs
+++ b/packages/blitz-paint/src/render/box_shadow.rs
@@ -7,14 +7,22 @@ use peniko::{Compose, Fill, Mix};
 impl ElementCx<'_> {
     pub(super) fn draw_outset_box_shadow(&self, scene: &mut impl PaintScene) {
         let box_shadow = &self.style.get_effects().box_shadow.0;
-
-        // TODO: Only apply clip if element has transparency
         let has_outset_shadow = box_shadow.iter().any(|s| !s.inset);
         if !has_outset_shadow {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


packages/blitz-paint/src/render/box_shadow.rs
```diff
@@ -7,13 +7,14 @@ use peniko::{Compose, Fill, Mix};
...
         let box_shadow = &self.style.get_effects().box_shadow.0;
-
-        // TODO: Only apply clip if element has transparency
         let has_outset_shadow = box_shadow.iter().any(|s| !s.inset);
         if !has_outset_shadow {
             return;
         }
 
+        let opacity = self.style.get_effects().opacity;
+        let needs_clip = opacity < 1.0;
```
nicoburns: I think this also needs to check that the background color is opaque

nicoburns (changes requested):

nicoburns (approved):
Seems reasonable. We really ought to measure that it's actually faster. But happy to accept this for now. It's easy to revert if necessary.

PR #401, opened 2026-04-25
fix: implement case-sensitive attribute selectors

lainon1:
## Summary
- Implement CSS `[attr="value" i]` case-insensitive attribute selector support
- Uses `Cow<str>` to avoid allocations in the common case (case-sensitive), only allocating when lowercase conversion is needed

## Changes
- `packages/blitz-dom/src/stylo.rs`: Added case sensitivity handling for all attribute selector operators (Equal, Includes, DashMatch, Prefix, Substring, Suffix)
- Removed the TODO comment

## Testing
- The code compiles successfully
- All 16 unit tests in blitz-dom pass
- Matches the CSS Selectors Level 4 spec for ASCII case-insensitive attribute matching

### DIFF
--- a/packages/blitz-dom/src/stylo.rs
+++ b/packages/blitz-dom/src/stylo.rs
@@ -11,7 +11,7 @@ use markup5ever::{LocalName, LocalNameStaticSet, Namespace, NamespaceStaticSet, 
 use selectors::bloom::BLOOM_HASH_MASK;
 use selectors::{
     Element, OpaqueElement,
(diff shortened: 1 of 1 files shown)


nicoburns (changes requested):
Looking into this, I think this entire function should be:

```rust
match self.data.attr(local_name.0.clone()) {
      None => false,
      Some(attr_value) operation.eval_str(attr_value),
}
```

PR #402, opened 2026-04-27
feat(browser): add opt-in FPS counter overlay

tonybierman:
## Summary
- Adds an opt-in FPS counter overlay to the Blitz browser UI (`apps/browser`), gated behind a `--show-fps` CLI flag.
- Frame timing is measured via a `dioxus_native::CustomPaintSource` whose `render` method fires once per real paint — i.e. the wall-clock delta between successive `RedrawRequested` events, so the displayed FPS reflects actual paint cadence.
- All changes are localized to `apps/browser/`; no engine packages were modified.

Addresses the "FPS counter" checklist item in #363.

## Design
- `apps/browser/src/fps_overlay.rs` (new) — `FpsStats` (60-frame ring of frame durations), `FpsPaintSource` (impl `CustomPaintSource`, records `Instant::now()` deltas in `render()`, resets on `suspend()`), and `FpsOverlay` (Dioxus component that registers the source via `use_wgpu`, polls the stats every 250ms, and renders a 1×1 tick canvas plus the visible HUD).
- `apps/browser/src/main.rs` — parses `--show-fps` from `std::env::args()` and threads a `ShowFps` newtype through `dioxus_native::launch_cfg`'s context vector. The overlay component is conditionally mounted in the chrome.
- `apps/browser/assets/browser.css` — `.fps-overlay` (fixed top-right, semi-transparent black, green monospace) and `.fps-tick` (1×1 invisible canvas that drives the per-frame tick). Note: the tick canvas uses `opacity: 0.01` rather than `0` because `blitz-paint`'s renderer short-circuits elements with `opacity: 0`, which would otherwise prevent `CustomPaintSource::render` from firing.
- `apps/browser/Cargo.toml` — adds `tokio = { workspace = true, features = ["time"] }` for the 250ms polling task (already in the workspace lockfile).

## Test plan
- [ ] `cargo build -p browser` — clean build.
- [ ] `cargo run -p browser` — overlay does NOT appear (no flag).
- [ ] `cargo run -p browser -- --show-fps` — overlay appears top-right, below the urlbar.
- [ ] Load an animated page (e.g. anything with CSS animation) — FPS rises to a steady value.
- [ ] Load a static page — FPS settles to ~0 within ~1s.
- [ ] Existing Alt+D / Alt+H devtools toggles still work.

### DIFF
--- a/apps/browser/src/fps_overlay.rs
+++ b/apps/browser/src/fps_overlay.rs
@@ -2,7 +2,7 @@ use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
 use dioxus_native::prelude::*;
-use dioxus_native::{use_wgpu, CustomPaintCtx, CustomPaintSource, DeviceHandle, TextureHandle};
+use dioxus_native::{CustomPaintCtx, CustomPaintSource, DeviceHandle, TextureHandle, use_wgpu};
 
 const RING_LEN: usize = 60;
 
--- a/apps/browser/src/fps_overlay.rs
+++ b/apps/browser/src/fps_overlay.rs
@@ -1,0 +1,116 @@
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use dioxus_native::prelude::*;
+use dioxus_native::{CustomPaintCtx, CustomPaintSource, DeviceHandle, TextureHandle, use_wgpu};
+
+const RING_LEN: usize = 60;
+
+struct FpsStats {
+    last: Option<Instant>,
+    ring: [Duration; RING_LEN],
+    idx: usize,
+    count: usize,
+}
+
+impl Default for FpsStats {
+    fn default() -> Self {
+        Self {
+            last: None,
+            ring: [Duration::ZERO; RING_LEN],
+            idx: 0,
+            count: 0,
+        }
+    }
+}
+
+impl FpsStats {
+    fn record(&mut self, now: Instant) {
+        if let Some(prev) = self.last {
+            let dt = now.duration_since(prev);
+            self.ring[self.idx] = dt;
+            self.idx = (self.idx + 1) % RING_LEN;
+            if self.count < RING_LEN {
+                self.count += 1;
+            }
+        }
+        self.last = Some(now);
+    }
+
+    fn snapshot(&self) -> (f32, f32) {
+        if self.count == 0 {
+            return (0.0, 0.0);
+        }
+        let total: Duration = self.ring[..self.count].iter().copied().sum();
+        let avg = total / self.count as u32;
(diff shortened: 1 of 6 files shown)


tonybierman:
Screenshot From 2026-04-27 08-39-00

apps/browser/src/main.rs
```diff
@@ -44,7 +48,14 @@ pub fn android_main(android_app: dioxus_native::AndroidApp) {
...
-    dioxus_native::launch(app)
+    let show_fps = std::env::args().any(|a| a == "--show-fps");
+    dioxus_native::launch_cfg(
+        app,
+        vec![Box::new(move || {
+            Box::new(ShowFps(show_fps)) as Box<dyn std::any::Any>
+        })],
+        Vec::new(),
+    )
 }
```
nicoburns: I'm not seeing any advantage to passing this in as context. I suggest either:

- Reading `std::env::args()` within the component directly
- Make the fps meter toggle a menu item

apps/browser/src/fps_overlay.rs
```diff
@@ -0,0 +1,116 @@
...
+        None
+    }
+}
+
+#[component]
+pub fn FpsOverlay() -> Element {
+    let stats: Arc<Mutex<FpsStats>> = use_hook(|| Arc::new(Mutex::new(FpsStats::default())));
+
+    let stats_for_source = stats.clone();
+    let paint_source_id = use_wgpu(move || FpsPaintSource {
```
nicoburns, line 90: As this is using `use_wgpu` which only exists with the `vello` rendering backend, this needs to be gated behind the `vello` feature.

nicoburns (changes requested):
We probably ought to build this into the framework, but I guess this `CustomPaintSource` hack is workable (albeit it restricts it to the Vello renderer).

tonybierman:
Updated based on review feedback:

- Removed the `--show-fps` command-line flag and `ShowFps` context — FPS is now toggled exclusively via the menu
- Added **Toggle FPS** menu item below **Toggle DevTools**
- Lowered FPS overlay `z-index` to 50 (below the dropdown menu at 100) so the menu always renders on top

tonybierman:
Good catch — fixed. `fps_overlay` module, the `show_fps` signal, the **Toggle FPS** menu item, and the overlay render are all now gated behind `#[cfg(feature = "vello")]`, consistent with the feature gating on `use_wgpu` in `dioxus-native`. Verified compiles cleanly with and without the `vello` feature.

nicoburns (approved):
Looks good now. Just needs rustfmt to be run :)

tonybierman:
Applied `cargo fmt` to reorder imports in `capture.rs`, `fps_overlay.rs`, and `main.rs`. No logic changes.

PR #403, opened 2026-04-27
Fix O(n²) complexity in collect_inline_roots_in_range

lainon1:
Changed Vec.contains() (O(n)) to HashSet.contains() (O(1)) for deduplication checks in collect_inline_roots_in_range, reducing overall complexity from O(n²) to O(n).

### DIFF
--- a/packages/blitz-dom/src/traversal.rs
+++ b/packages/blitz-dom/src/traversal.rs
@@ -1,5 +1,4 @@
 use std::cmp::Ordering;
-use std::collections::HashSet;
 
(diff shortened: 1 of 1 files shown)


nicoburns:
Hmm... I think we ought to use traversal logic that only ever visits each node once by construction. Then we could just check whether we've reached the end node yet, and wouldn't need to keep track of visited nodes at all.

lainon1:
Updated implementation - now starts traversal from first_anchor instead of root, eliminating the 
need for any visited tracking by construction.

PR #406, opened 2026-04-28
feat(browser): tabs (multiple embedded documents)

tonybierman:
## Summary

Implements the following features from the Browser UI backlog (#363):

- Tabs
- Status bar
- Menu item to clear HTTP cache

It also refactors the Browser UI into multiple files.

Out-of-scope items deliberately deferred for follow-ups:
- drag-to-reorder tabs
- `target=_blank` → new tab
- keyboard shortcuts (Ctrl+T/W/Tab),
- pre-warmed hidden viewport (layout hidden tabs with correct viewport)
- pinning / right-click menu / thumbnails, session persistence.

Implementation details: 

- One `<web-view>` per tab, rendered with `display: none` on inactive tabs so style resolution and asset fetching continue in the background while paint is skipped.
- Per-tab `Tab` struct: `history`, `DocumentLoader`, `document`, `node_handle`, `html_source`, `title`. Each tab loads independently; closing a tab drops its loader and cancels in-flight fetches.
- Tab strip with click-to-switch, "+" new-tab, "×" close (hidden when only one tab remains). Rounded upper corners, seamless join with the urlbar.
- Live tab titles via `find_title_node()` cached into a per-tab Signal at load time (falls back to URL).
- Window title reactively follows the active tab's title via the chrome's `<title>` element.
- Back/forward buttons render disabled (dim, click is a no-op) when there's no history in that direction.
- All chrome handlers (back/forward/refresh/home/menu/URL bar) operate on the active tab read at call time.


tonybierman:
Screenshot From 2026-04-27 20-12-31

tonybierman:
@nicoburns  I was going to move components out of main to their own code files

nicoburns:
@tonybierman Seems reasonable. I've noticed a couple of bugs:

- ~~Documents weren't unloading properly when a tab was closed~~ (this was a bug in `blitz-dom`: I've pushed a fix for this)
- Navigation only works in the right-most tab (regardless of which tab is active). If I submit a url in the url bar or click a link in a non-rightmost tab then the url in the url bar updates, but the page only actually loads if I press the "refresh" button.
- The browser seems to be stuck in "continually rendering frames" mode. Normally this is supposed to happen when there is an active animation. I can't see one, but perhaps there's one I can't see. Could potentially also be linked to https://github.com/DioxusLabs/blitz/pull/408 (perhaps we're not removing animation correctly if the DOM node gets removed before the animation completes?)

tonybierman:
> @tonybierman Seems reasonable. I've noticed a couple of bugs:
> 
> * ~Documents weren't unloading properly when a tab was closed~ (this was a bug in `blitz-dom`: I've pushed a fix for this)
> * Navigation only works in the right-most tab (regardless of which tab is active). If I submit a url in the url bar or click a link in a non-rightmost tab then the url in the url bar updates, but the page only actually loads if I press the "refresh" button.
> * The browser seems to be stuck in "continually rendering frames" mode. Normally this is supposed to happen when there is an active animation. I can't see one, but perhaps there's one I can't see. Could potentially also be linked to [Fix panic in resolve_stylist when animation entry is stale #408](https://github.com/DioxusLabs/blitz/pull/408) (perhaps we're not removing animation correctly if the DOM node gets removed before the animation completes?)

I'll take a look

nicoburns:
> The browser seems to be stuck in "continually rendering frames" mode. Normally this is supposed to happen when there is an active animation. I can't see one, but perhaps there's one I can't see. Could potentially also be linked to https://github.com/DioxusLabs/blitz/pull/408 (perhaps we're not removing animation correctly if the DOM node gets removed before the animation completes?)

This was the status bar code unconditionally writing to a Dioxus signal even if the value hadn't changed. The Signals model of state that Dioxus uses is that any write triggers an update (even if they value is the same). To avoid rerenders you have to compare and write conditionally.

(fix pushed)

tonybierman:
> > The browser seems to be stuck in "continually rendering frames" mode. Normally this is supposed to happen when there is an active animation. I can't see one, but perhaps there's one I can't see. Could potentially also be linked to #408 (perhaps we're not removing animation correctly if the DOM node gets removed before the animation completes?)
> 
> This was the status bar code unconditionally writing to a Dioxus signal even if the value hadn't changed. The Signals model of state that Dioxus uses is that any write triggers an update (even if they value is the same). To avoid rerenders you have to compare and write conditionally.
> 
> (fix pushed)

I merged your fix into my refactor (components out of main) and pushed.  Haven't worked on bugs from your previous comment yet, will do shortly.

nicoburns:
> Navigation only works in the right-most tab (regardless of which tab is active). If I submit a url in the url bar or click a link in a non-rightmost tab then the url in the url bar updates, but the page only actually loads if I press the "refresh" button.

Just a note for clarity that this is the only remaining bug from that comment (fixes pushed for the other - but I'm not planning on looking at this one right now)

nicoburns:
Tested on mobile, and it actually works pretty well! Only thing is that the tab bar should probably be a (horizontal) scroll container, because you run out of space for more tabs very quickly:

https://github.com/user-attachments/assets/79f1af07-a6c0-455a-af64-1d09408bbf07

nicoburns:
Running on ios sim with `dx serve --ios -p browser --no-default-features --features hybrid,incremental,floats --release` btw. Android also works although I'd recommend then`skia` renderer rather than the `hybrid` renderer on Android. `cargo install dioxus-cli` to get `dx`.

tonybierman:
@nicoburns I handled the tab abundance issue by progressively compressing each tab's width as new tabs are added.  I feel like we're at a good stopping point for this PR if you feel like it is ready.

nicoburns:
I'm mostly happy with where this is (a lot of new features!). And I'm keen to avoid this PR getting too much bigger. However, your fix for the navigation issue has a problem: now the page is reloaded every time the active tab is changed! This is a little tricky to get right. I *think* the easist way to fix this may be to put the effect that reloads on history change into a component that only has access to one Tab's state. That feels a little hacky, but I'm not currently able to think of a better way.

packages/blitz-dom/src/resolve.rs
```diff
@@ -186,6 +192,14 @@ impl BaseDocument {
...
                 let layout_children = doc.nodes[node_id].layout_children.borrow_mut().take();
                 if let Some(layout_children) = layout_children {
+                    // Filter out stale IDs — anonymous blocks and pseudo-elements
+                    // can be removed from the slab between render passes, so cached
+                    // layout_children may contain keys that are no longer valid.
+                    let layout_children: Vec<usize> = layout_children
+                        .into_iter()
+                        .filter(|&id| doc.nodes.contains(id))
+                        .collect();
+
```
nicoburns: We should avoid allocating another `Vec` here, and instead put the `contains` check in the loop just below, and `continue` from the loop if the node doesn't exist.

nicoburns (changes requested):

nicoburns (approved):
This LGTM to me now. And I've cleaned up the sloppy PR description.

PR #408, opened 2026-04-28
Fix panic in resolve_stylist when animation entry is stale

### LINKED ISSUE #407: Panic in `resolve_stylist`: "invalid key" when animation node tracked but removed
**Version**: blitz 0.3.0-alpha.2 (blitz-dom)
**Platform**: Linux (Vulkan/Wayland), likely cross-platform — bug is in Slab indexing logic.

### Repro

```rust
use blitz_dom::{DocumentConfig, Document as _};
use blitz_html::HtmlDocument;

fn main() {
    let html = r#"
        <style>
          @keyframes pulse { 0% { opacity: 1; } 100% { opacity: 0.5; } }
          .pulse { animation: pulse 2s infinite; }
        </style>
        <div id="pulse-node" class="pulse">animated</div>
    "#;

    let mut doc = HtmlDocument::from_html(html, DocumentConfig::default());

    // First resolve — animation entry is registered for #pulse-node in animations.sets.
    doc.as_mut().resolve(0.0);

    // Remove the animated element (its slab key is now freed).
    let pulse_id = doc.inner().get_element_by_id("pulse-node");
    if let Some(id) = pulse_id {
        doc.inner_mut().mutate().remove_and_drop_node(id);
    }

    // Second resolve — panics: animations.sets still references the freed slab key.
    doc.as_mut().resolve(0.1); // panics here
}
```

Any DOM removal of an element with an active CSS `animation:` rule followed by a second `resolve()` reproduces it. The same panic also occurs in the GUI event loop when blitz internally recreates nodes between frames.

### Panic

```
thread 'main' panicked at blitz-dom-0.3.0-alpha.2/src/stylo.rs:83:23:
invalid key
```

### Root Cause

In `resolve_stylist()` (around stylo.rs:83):

```rust
self.nodes[node_id].set_restyle_hint(RestyleHint::RESTYLE_SELF);
```

Flow:
1. **First resolve** — `traverse_dom` matches CSS animations against the DOM,
   writes `node_id`s into `animations.sets`.
2. The animated node is removed from the DOM (slab key freed).
3. **Second resolve** — iterates `animations.sets` and indexes `self.nodes` directly.
   The slab key is stale, so the bracket index panics.

### Proposed Fix

Use `get_mut` and skip stale entries:

```rust
let Some(node) = self.nodes.get_mut(node_id) else {
    continue; // stale animation entry — skip safely
};
node.set_restyle_hint(RestyleHint::RESTYLE_SELF);
```

Vendored locally with this patch — reproducer passes cleanly, GUI event loop runs cleanly through hundreds of frames.

### PR

Happy to submit a PR with the 1-line fix + regression test if it matches your design intent.

---

_Drafted with Claude, fix and reproduction hand-verified._

yes1688:
## Summary

Fixes #407 — panic with `"invalid key"` in `resolve_stylist` when an animation
entry's Slab key is stale after node removal.

## Change

1-line fix using `get_mut` to skip stale entries safely:

```rust
let Some(node) = self.nodes.get_mut(node_id) else {
    continue; // stale animation entry — node removed from slab
};
node.set_restyle_hint(RestyleHint::RESTYLE_SELF);
```

## Verification

- Reproduction in #407 panics on `main` (verified locally)
- With this patch, the same reproduction runs cleanly through hundreds of frames
- Used in a downstream Rust app (HTML console) for ~1 day with no regressions
- Regression test added in `packages/blitz-html/tests/html_document_test.rs`

## Notes

If you'd prefer a different approach (e.g. cleaning up `animations.sets` on node
removal at the source), happy to rework — just let me know your preference.

---

_Drafted with Claude, fix and reproduction hand-verified._

### DIFF
--- a/packages/blitz-dom/src/stylo.rs
+++ b/packages/blitz-dom/src/stylo.rs
@@ -76,11 +76,17 @@ impl crate::document::BaseDocument {
             .flush(&guards)
             .process_style(root, Some(&self.snapshots));
 
-        // Mark actively animating nodes as dirty
+        // Mark actively animating nodes as dirty.
+        // Use get_mut to skip stale entries: stylo can hold animation keys for
+        // nodes already removed from the slab, and direct indexing panics on
+        // invalid keys. See https://github.com/DioxusLabs/blitz/issues/407
         let mut sets = self.animations.sets.write();
         for (key, set) in sets.iter_mut() {
             let node_id = key.node.id();
-            self.nodes[node_id].set_restyle_hint(RestyleHint::RESTYLE_SELF);
+            let Some(node) = self.nodes.get_mut(node_id) else {
+                continue; // stale animation entry — node removed from slab
+            };
+            node.set_restyle_hint(RestyleHint::RESTYLE_SELF);
 
             for animation in set.animations.iter_mut() {
                 if animation.state == AnimationState::Pending && animation.started_at <= now {
--- a/packages/blitz-dom/tests/stylo_test.rs
+++ b/packages/blitz-dom/tests/stylo_test.rs
@@ -1,0 +1,9 @@
+use blitz_dom::{BaseDocument, DocumentConfig};
+
+/// Smoke-test: resolve on an empty document must not panic.
+#[test]
+fn resolve_empty_document_does_not_panic() {
+    let mut doc = BaseDocument::new(DocumentConfig::default());
+    doc.resolve(0.0);
+    doc.resolve(0.1);
+}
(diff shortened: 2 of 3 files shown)


nicoburns:
Thanks for the PR. I've implemented a similar but slightly modified fix in https://github.com/DioxusLabs/blitz/commit/c5e90040bac430cbc6f5de5f48158aad7b2d7ab0. I used your test and credited you as co-author.

nicoburns:
Sorry for the delay!

PR #410, opened 2026-04-28
feat(browser): special pages for in-app routes (Settings, History, Bookmarks, Start)

tonybierman:
## Summary

Generalizes the existing `about:newtab` handling into a registry-based dispatcher for "special pages" — in-app routes served on the `about:` scheme and rendered by Rust functions that emit HTML strings on each navigation. State lives in the URL plus an in-memory `ConfigStore`, so links/forms drive UI updates through the existing `HtmlDocument` pipeline (no live `VirtualDom` per tab; matches the precedent set by `about:newtab` and `view_source_action`).

Initial pages:
- `about:newtab` — replaces `assets/start.html` with a Dioxus-rendered start page that links to the other special pages
- `about:settings` — theme toggle (light/dark) persisted in `ConfigStore` for the session
- `about:history` — read-only listing of the current tab's history (cross-tab history flagged as a follow-up)
- `about:bookmarks` — placeholder

Menu entries for Settings/History/Bookmarks added to the toolbar dropdown. The URL bar already accepts `about:` URLs via `Url::parse`, so typing `about:settings` works without changes to `req_from_string`.

## Architecture

- `apps/browser/src/special_pages/mod.rs` — `SpecialPage` trait, static registry, `lookup`/`dispatch`, shared CSS + page shell.
- `apps/browser/src/special_pages/{start,settings,history_page,bookmarks}.rs` — one page per file.
- `apps/browser/src/config.rs` — small `Arc<RwLock<HashMap>>` config store; on-disk persistence intentionally deferred.
- `document_loader.rs` — `about:` branch now dispatches through the registry; `base_url` is set so action links (`about:settings/set?key=...&value=...`) resolve through `BrowserNavProvider`. `<title>` is parsed for special pages (was previously empty for `about:newtab`).
- `tab.rs` / `main.rs` — `Arc<ConfigStore>` threaded through `Tab::new` → `DocumentLoader::new`.

## Notes

- `assets/start.html` is now dead code; left in this PR for reviewer continuity but a follow-up commit can delete it.
- Settings persistence is in-memory only for milestone 1 — values reset on relaunch. On-disk JSON persistence under `dirs::config_dir()/blitz/` is the natural follow-up.
- `about:` URLs are non-special per WHATWG; this PR sidesteps relative-resolution quirks by always emitting fully-qualified hrefs from special pages. Falling back to a `blitz://` scheme is a one-line change in `lookup` if needed.

Refs #363

## Test plan

- [ ] Launch: `cargo run --release -p browser` opens to the new start page (logo + Settings/History/Bookmarks links).
- [ ] Menu items: clicking Settings / History / Bookmarks navigates to the corresponding `about:` URL.
- [ ] URL bar: typing `about:settings` (or `about:history`, etc.) and pressing Enter loads the page.
- [ ] Settings action: on `about:settings`, clicking "Switch to dark" navigates to `about:settings/set?key=theme&value=dark`, page re-renders dark; "Switch to light" toggles back.
- [ ] Back/forward through `about:newtab → about:settings → about:settings/set?...` works.
- [ ] View Source on a special page shows the SSR-rendered HTML.
- [ ] Existing tab behavior (HTTP fetch, error fallback, tab strip, view-source for HTTP pages) is unchanged.


tonybierman:
@nicoburns for your review, per our conversation

apps/browser/src/special_pages/blitz_logo.b64
```diff
@@ -0,0 +1 @@
...6fObpeFELu4JzIrnAcesIzMKr8qhQOvExGrqEqyEq0XexYscaix025e5kwMstsraud3XAc8zVj/nc6lHgMfWXZkZ7c4U8IgxgQcHImAyA5kJA8MBTFmaB6PWUS0CLkejf5V3VQVwZWj5USq+J4eAzTYDmjd1A2MG2MS9uMCP+IEI+BSV1jkLX7cBX5K9ZvtyJ/eABLNDX7lbmgMkNFiZl1jd7gMgQtGpyt3yTCBCKgs8OioCKkpznyeO7ZD+kVc0zzOAg7LcCSYOYp6nHdx18VgaiFFMz5Jvi2uYNhCTo+21AbJOEkDSTPU3X/CDQFWAILG5K/BtXuyEKPpGC97VpkDWPV56cMs8eQSkObrhhl/xbd8wFSBMzLN/vbKSUgHIU0zNH8hhGQRa3wbyanS9FxWm/QaEo696xJO3AkN3IBQSHzdqq6omITSKqXlEVvu262mmAqHJ8Ho2XyT1UoQw2VNd84ILtOwHbsdQTQdClawL0XMnAxug9HXN6/iu1f5/1JYb/A57akP40nnez2kLVHOwQYrTN3VV1TTDMDyvEwSB53mGoWmaqurmTIFNqrA+BoduJ1aAKErwdoe6nv2mCFFTLPPbaKfq004PscnftTmCeg6io0LTi+23KV5KQjRk+L5O28Z+UyYKVyvr2CnzHvuLqrB18/xWBH7V2P5NYnh2uY09sY18sVSMx0U+s13Yl4vXE8AVsUL3E86UuF/hp0snNaN2fYrs1YM08CB7sO2+rWH/gP09Xa0UkfE3cnZLNWBYurHtxCC505SATck9Nl+Bo0KqzOBePt2I9nU5AY/3ssCSRIyfxyRu0c7Dp6xs5pP56Myxhy5Vf0T/2b3woBr5kRjCjmOHNOcuNY62B7QwHJcprXepUo/oQPPP9u1tB0EYBgNwVgUJGQM5g4AOjMpBFOT9n807b00MjIH9XqHpuj/rhDDTSrYYp5U2nupTU2kgzVMc8RkuQImiA5+/4bUkw3Qmlmo/ZrzZxdUR11rnsaMWFx/hYw4OTvFZqRRydxCEKG2PHS6JJguMeJiWd7UoBjPZNCGrPDKMb+snL4rrEPJS9QvLlbEG/eGeQPG/v4oXJqIpqw2N/NrbCm8h07G7l2gTOWFnlbmxd8n3SmsKrwPoiycWey120dmhYZZ2AOzDAkj7sKCnm4kpDK3YG/ax+Rblicg0AAAAAElFTkSuQmCC
```
nicoburns: This should be stored as a regular PNG/SVG (or whatever kind of file it is, and then loaded as an asset using Dioxus's asset system. See `icons.rs` for how to do this.

nicoburns:
@tonybierman Full review coming, but I see you're doing more stuff with light mode / dark mode and wanted to note that Blitz has CSS support for this. If we can set the theme on the winit window then we should be able to use a regular media query to theme the browser, and this should also make dark mode work for actual website content!

tonybierman:
> @tonybierman Full review coming, but I see you're doing more stuff with light mode / dark mode and wanted to note that Blitz has CSS support for this. If we can set the theme on the winit window then we should be able to use a regular media query to theme the browser, and this should also make dark mode work for actual website content!

Ok, I'll halt on the theming stuff until I learn more about that.

packages/blitz-shell/Cargo.toml
```diff
@@ -32,6 +32,7 @@ anyrender = { workspace = true }
 
 # Windowing & Input
 winit = { workspace = true }
+dark-light = { workspace = true }
```
nicoburns: Can we ditch the `dark-light` dependency? (unless you have strong feelings about this). It brings in a lot of dependencies for only a small increase in functionality (and I don't think this will correctly "subscribe to changes"). If `dark-light` is detecting the theme in cases where Winit doesn't, then we should ideally look at getting this fixed in Winit?

apps/browser/Cargo.toml
```diff
@@ -5,6 +5,13 @@ edition = "2024"
 license.workspace = true
 publish = false
 
+[lints.clippy]
+unwrap_used             = "warn"
+expect_used             = "warn"
+panic                   = "warn"
+indexing_slicing        = "warn"
+arithmetic_side_effects = "warn"
+
```
nicoburns: Why have these been added? Can we remove please?

apps/browser/src/icons.rs
```diff
@@ -1,18 +1,36 @@
...
+        match ch {
+            '#' => out.push_str("%23"),
+            '\n' | '\r' | '\t' => out.push(' '),
+            '"' => out.push_str("%22"),
+            '<' => out.push_str("%3C"),
+            '>' => out.push_str("%3E"),
+            ' ' => out.push_str("%20"),
+            _ => out.push(ch),
+        }
+    }
```
nicoburns, line 31: If we're going to do this then we should use a crate for the escaping. Perhaps it would be best to use base64 encoding?

apps/browser/src/special_pages/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub url: &'a Url,
+    pub history: SyncStore<History>,
+    pub config: Arc<ConfigStore>,
+}
+
+pub trait SpecialPage: Send + Sync {
+    fn host(&self) -> &'static str;
+    fn render(&self, ctx: &SpecialPageCtx<'_>) -> String;
+    fn handle_action(&self, _ctx: &SpecialPageCtx<'_>) {}
+}
```
nicoburns: I don't like that these special pages are still string-based. I think we ought to modify `Tab` so that it has an enum as a field that allows us to choose between rendering a `<web-view>` with a sub document, or rendering a Dioxus component inline. Something like:

```rust
enum TabContent {
    ContentPage(SubDocumentContent), // Existing SubDocument based tab
    SpecialPage(SpecialPageContent), // Dioxus component based tab
}
```

In the Dioxus component case, there wouldn't be a sub document with an separate VirtualDom. The Dioxus component would just be directly rendered into the main Browser UI's (Virtual) DOM.

`SpecialPageContent` could probably hold something like a boxed closure than returns a `dioxus::Element`.

apps/browser/src/icons.rs
nicoburns: The alternative would be to make the SVGs inline SVGs. Then `currentColor` will work.

nicoburns (changes requested):
The dark mode CSS is looking good now. It would be nice to have the option to set an override theme via settings (Winit exposes this option, but `dioxus-native`/`blitz-shell` don't currently expose it to Dioxus application code: we probably need to a `DesktopContext` type like `dioxus-desktop` has). But that's probably best left as a follow-up.

See my comments about making the special pages Dioxus components rather than HTML sub documents.

nicoburns:
Also: a feature request. Not technically related to this PR, but it's small so it could be included: could we get a search box on the new tab page? It could submit directly to Duck Duck Go search like the urlbar does. I think it should be possible for this to just be an HTML form.

apps/browser/src/special_pages/start.rs
```diff
@@ -1,18 +1,21 @@
...
     fn host(&self) -> &'static str {
         "newtab"
     }
 
     fn render(&self, _ctx: &SpecialPageCtx<'_>) -> String {
+        let logo_url = format!("file://{}", BLITZ_LOGO.resolve().display());
         let body = format!(
             r#"<div style="display:flex;flex-direction:column;align-items:center;justify-content:center;min-height:60vh;text-align:center;">
-  <img src="data:image/png;base64,{LOGO_B64}" alt="Blitz" style="width:160px;height:160px;">
+  <img src="{logo_url}" alt="Blitz" style="width:160px;height:160px;">
```
nicoburns: You should just be able to do `src={BLITX_LOGO}` here

nicoburns: Oh... you would be able to if this was a Dioxus component. I guess that won't work in string.

packages/blitz-shell/src/application.rs
```diff
@@ -6,7 +6,35 @@ use std::sync::mpsc::Receiver;
...
+    target_os = "netbsd",
+    target_os = "openbsd",
+))]
+fn spawn_theme_watcher(proxy: BlitzShellProxy) {
+    std::thread::Builder::new()
+        .name("theme-watcher".into())
+        .spawn(move || {
+            let mut last_mode = dark_light::detect().ok();
+            loop {
+                std::thread::sleep(std::time::Duration::from_secs(1));
```
nicoburns: Every second seems a bit much for functionality that will probably hardly ever get hit. Can we use https://docs.rs/ashpd/latest/ashpd/desktop/settings/struct.Settings.html#method.receive_color_scheme_changed for this. I don't really like the `ashpd` dependency as it's kinda huge for what it does. But `light-dark` depends on it anyway so we might as well use it here.

tonybierman: the polling approach was reverted

apps/browser/src/history.rs
```diff
@@ -34,29 +37,41 @@ impl<Lens> Store<History, Lens> {
...
+            *self.current().write() = self.current_idx().saturating_sub(1);
         }
     }
 
     fn go_forward(&mut self) {
         if self.has_forward() {
-            *self.current().write() += 1;
+            *self.current().write() = self.current_idx().saturating_add(1);
         }
     }
```
nicoburns, line 54: Can we undo these please? If the history overflows `usize::MAX` then we have bigger problems! And the saturating code is much less readable.

apps/browser/Cargo.toml
nicoburns: I think you asked why earlier: because they have bunch of false positives, and the `allow`s add a lot of noise to the code for not a lot of value.

apps/browser/assets/browser.css
```diff
@@ -292,3 +313,267 @@ body,
   color: #333;
   z-index: 10;
 }
+
+#frame.dark {
+  color: #e6e6e6;
+}
```
nicoburns, line 319: Not blocking, but could we use CSS variables to reduce the duplication here? It seems very unfortunate to have duplicate declarations for so many different parts of the UI.

apps/browser/src/special_pages/history_page.rs
```diff
@@ -0,0 +1,65 @@
...
+        .map(|(req, title)| {
+            let url = req.url.clone();
+            let display = title
+                .as_deref()
+                .filter(|t| !t.trim().is_empty())
+                .unwrap_or(url.as_str())
+                .to_string();
+            (url, display)
+        })
+        .collect();
```
nicoburns, line 30: Can we wrap this in a [`use_memo`](https://docs.rs/dioxus/latest/dioxus/prelude/fn.use_memo.html) to avoid unnecessary recomputations?

apps/browser/src/special_pages/history_page.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::config::ConfigStore;
+use crate::history::{History, HistoryStoreExt, SyncStore};
+
+pub fn render(
+    history: SyncStore<History>,
+    _config: Arc<ConfigStore>,
+    navigate: NavigateFn,
+) -> Element {
+    let urls: Vec<_> = history.urls().read().clone();
+    let titles: Vec<_> = history.titles().read().clone();
```
nicoburns, line 15: These should not need to be cloned. You should be able to iterate over the borrowed contents.

apps/browser/src/special_pages/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+mod bookmarks;
+mod history_page;
+mod settings;
+mod start;
+
+/// Callback passed to special-page render functions so they can trigger tab navigation.
+pub type NavigateFn = Arc<dyn Fn(Url) + Send + Sync>;
+
+type RenderFn = fn(SyncStore<History>, Arc<ConfigStore>, NavigateFn) -> Element;
```
nicoburns, line 17: Rather than passing these variables in here, perhaps we should provide them via [context](https://docs.rs/dioxus/latest/dioxus/prelude/fn.use_context_provider.html)?

(context is keyed by type)

apps/browser/src/tab.rs
```diff
@@ -25,7 +27,9 @@ pub fn next_tab_id() -> TabId {
 pub struct Tab {
     pub id: TabId,
     pub history: SyncStore<History>,
+    pub load_trigger: LoadTriggerSignal,
     pub loader: Rc<DocumentLoader>,
+    pub content: Signal<TabContent>,
     pub document: Signal<Option<SubDocumentAttr>>,
     pub node_handle: Signal<Option<NodeHandle>>,
     pub html_source: Signal<String>,
```
nicoburns, line 35: `document`, `node_handle`, and `html_source` seem like they should like inside `TabContent::Web`. As I believe they only apply to web tabs? What do you think?

apps/browser/src/tab.rs
```diff
@@ -65,6 +78,7 @@ impl Tab {
 
 pub fn active_tab(tabs: &Signal<Vec<Tab>>, active_id: TabId) -> Tab {
     let tabs_ref = tabs.read();
+    #[allow(clippy::expect_used)] // tabs is never empty: always initialised with one tab
```
nicoburns, line 81: These could be cleaned up now that we've removed the lint from the config.

apps/browser/src/document_loader.rs
```diff
@@ -5,13 +5,55 @@ use std::sync::{
...
+use crate::special_pages::{self, NavigateFn, SpecialPageComponent, TabContent};
+
+/// Drives the `use_effect` in `TabView`.
+///
+/// `NewNav` — user-initiated navigation; commit the resolved URL to history on success.
+/// `BackForward` — navigating within already-committed history; just reload and update title.
+#[derive(Clone)]
+pub enum LoadTrigger {
+    NewNav(Request),
+    BackForward(Request),
```
nicoburns, line 32: Seeing as the variants both contain the same content, perhaps this should be:

```rust
struct LoadRequest {
     request: Request,
     save_to_history: bool,
}
```

(save to history could also be a plain enum)

That might be clearer and easier to use than this enum?

apps/browser/src/special_pages/history_page.rs
```diff
@@ -0,0 +1,65 @@
...
+                ul {
+                    for (url, display) in entries {
+                        {
+                            let nav = navigate.clone();
+                            let url_clone = url.clone();
+                            rsx!(
+                                li { key: "{url}",
+                                    span {
+                                        class: "sp-link",
+                                        onclick: move |_| nav(url_clone.clone()),
```
nicoburns, line 48: This clone seems unecessary. You already cloned above.

apps/browser/src/capture.rs
```diff
@@ -102,7 +118,7 @@ fn render_scene(
 pub(crate) async fn try_get_save_path(file_type_name: &str, ext: &str) -> Option<PathBuf> {
     let timestamp = std::time::SystemTime::now()
         .duration_since(std::time::UNIX_EPOCH)
-        .unwrap()
+        .unwrap_or_default()
         .as_secs();
     let default_name = format!("blitz-screenshot-{timestamp}.{ext}");
```
nicoburns, line 123: These (the whole file) seem like drive-by changes unrelated to the PR. Can we undo them please. Crashing isn't good, but silently failing isn't either! I would consider these kind of changes in a standalone PR where they can be reviewed separately.

But tbh they also seem like the sort of code that LLMs think are good idea, but leads to verbose and hard to maintain codebases.

apps/browser/src/config.rs
```diff
@@ -0,0 +1,42 @@
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex, RwLock};
+
+type Listener = Box<dyn Fn(&str, &str) + Send + Sync>;
+
+#[derive(Default)]
+pub struct ConfigStore {
+    inner: RwLock<HashMap<String, String>>,
+    listeners: Mutex<Vec<Listener>>,
+}
```
nicoburns, line 10: Non-blocking, but I think we might be better off with one `Mutex` around everything here. I wouldn't expect a lot of concurrent reads? (would you?)

apps/browser/src/document_loader.rs
nicoburns: The methods below wouldn't be needed if this was  a struct.

apps/browser/src/special_pages/settings.rs
```diff
@@ -0,0 +1,43 @@
+use std::sync::Arc;
+
+use dioxus_native::prelude::*;
+
+use super::NavigateFn;
+use crate::config::ConfigStore;
+use crate::history::{History, SyncStore};
+
+pub fn render(
```
nicoburns, line 9: I think these special page functions need to become Dioxus `Component`s in order for it to be valid to call hooks inside them. That should be quite simple: add `#[component]` to each function (I think they have the right signature anyway!).

If we want to keep the arguments here then I think we'd need to manually create a struct annotated with `#[props]` for them (so all pages have the same type for the props). But I think we may want to switch this to use context anyway (see other comment)

nicoburns: Currently navigating to the settings page from another "special page" panics with:

thread 'main' (25643078) panicked at /Users/nico/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dioxus-core-0.7.4/src/scope_context.rs:387:9:
Unable to retrieve the hook that was initialized at this index.
                    Consult the `rules of hooks` to understand how to use hooks properly.

                    You likely used the hook in a conditional. Hooks rely on consistent ordering between renders.
                    Functions prefixed with "use" should never be called conditionally.

apps/browser/src/fps_overlay.rs
```diff
@@ -37,6 +39,8 @@ impl FpsStats {
         self.last = Some(now);
     }
 
+    #[allow(clippy::indexing_slicing, clippy::arithmetic_side_effects)]
+    // count <= RING_LEN and is checked > 0 before use
     fn snapshot(&self) -> (f32, f32) {
```
nicoburns, line 44: These comments could be removed

apps/browser/src/icons.rs
```diff
@@ -1,18 +1,40 @@
...
```
nicoburns, line 31: Could we use the `percent-encoding` crate for this escaping please? We already have it as a transitive dependency so might as well use it here too.

apps/browser/src/main.rs
```diff
@@ -85,19 +111,54 @@ fn main() {
...
+    let theme = use_signal_sync(|| config_store.get("theme").unwrap_or_else(|| "light".into()));
+    use_hook(|| {
+        let cs = config_store.clone();
+        cs.subscribe(move |key, value| {
+            if key == "theme" {
+                let mut theme = theme;
+                theme.set(value.to_string());
+            }
+        });
+    });
```
nicoburns, line 128: I think if we made the `ConfigStore` an actual Dioxus `Store` using the `#[store]` macro (see https://docs.rs/dioxus-stores/latest/dioxus_stores/attr.store.html) then we could directly subscribe to a "projection" of the store (e.g. a specific key, accessed via a method) using Dioxus's built in subscription logic for stores.

We should also be able to get rid of the Mutex/RwLock in ConfigStore, as `Store` already provides interior mutability.

apps/browser/src/main.rs
```diff
@@ -85,19 +111,54 @@ fn main() {
...
+        }
+    });
+    use_effect(move || {
+        let mut icon_color = icon_color;
+        icon_color.set(if theme() == "dark" {
+            "#e6e6e6"
+        } else {
+            "#1a1a1a"
+        });
+    });
```
nicoburns, line 144: This could just be:

```rust
let icon_color = use_signal(|| {
     if theme.read() == "dark" {
         "#e6e6e6"
      } else {
            "#1a1a1a"
      }
});
```

(could also be `use_signal_sync` is you need a `Sync` signal, but I don't think you do. Dioxus is inherently single-threaded, so you shouldn't need `Send`/`Sync` unless you are synchronising with something external).

The whole point of signals is that they automatically subscribe to each other (but you are opting out of that by using `peek` instead of `read`).

apps/browser/src/toolbar.rs
```diff
@@ -51,21 +54,28 @@ pub fn Toolbar(
...
-        tab.loader.load_document(request);
+        let mut lt = tab.load_trigger;
+        lt.set(LoadTrigger::BackForward(request));
     });
 
     let back_action = use_callback(move |_| {
-        active_tab(&tabs, *active_tab_id.peek()).history.go_back();
+        let mut tab = active_tab(&tabs, *active_tab_id.peek());
+        tab.history.go_back();
+        let req = (*tab.history.current_url().read()).clone();
```
nicoburns, line 64: I believe this can just be `tab.history.current_url()()`

apps/browser/src/config.rs
nicoburns: Update: see other comment about replacing this with a dioxus `Store` which would allow us to remove the Mutex/RwLock and the custom subscribing logic.

nicoburns (changes requested):
Having the theme stay as a local config for now seems reasonable. We can sync with the global setting as a follow-up.

I have left some specific comments:
- The most important comments are about making the special pages an actual `Component` and making the `ConfigStore` an actual `Store`.
- A lot of the are about removing slop / keeping the code concise. Some self-review in this area would be appreciated. The time it takes me to review thing is generally proportional to amount of code changes. So unnecessary code changes are unwelcome. And same goes for maintaining a codebase, and the total lines of code in that codebase in general.

tonybierman:
> Having the theme stay as a local config for now seems reasonable. We can sync with the global setting as a follow-up.
> 
> I have left some specific comments:
> 
> * The most important comments are about making the special pages an actual `Component` and making the `ConfigStore` an actual `Store`.
> * A lot of the are about removing slop / keeping the code concise. Some self-review in this area would be appreciated. The time it takes me to review thing is generally proportional to amount of code changes. So unnecessary code changes are unwelcome. And same goes for maintaining a codebase, and the total lines of code in that codebase in general.

I appreciate your feedback.  I think I should have communicated with you better up front; this was just a proof-of-concept I wanted your feedback on.  I didn't clean up much code while I was iterating. In any case, I'm going to close this draft PR and possibly revisit the idea at a future time.

PR #412, opened 2026-05-01
feat(browser): Dioxus Tabs

tonybierman:
## Summary
- Replace inline `about:newtab` HTML with a Dioxus-rendered `AboutPage` component.
- Extract URL/nav helpers from `toolbar.rs` into a new `nav` module.

## Stubs
`Settings`, `History`, and `Bookmarks` are wired through the menu to
Dioxus tabs and currently open a stub placeholder.


apps/browser/src/tab.rs
```diff
@@ -131,26 +141,55 @@ pub fn TabWebView(tab: Store<Tab>, active_tab_id: Signal<TabId>) -> Element {
...
 
+    // Title ownership: about pages set their own title here (no doc load happens
+    // for them); the memo de-dupes so this only fires on actual page changes.
+    // For real URLs, `apply_loaded_document` sets the title from the network
+    // response. Stale titles persist briefly during navigation, matching
+    // standard browser behavior.
+    use_effect(move || {
+        if let Some(page) = about() {
+            *tab.title().write_unchecked() = page.title().to_string();
+        }
```
nicoburns, line 158: This could probably go directly in the `use_memo` above. But this is also fine.

apps/browser/src/tab.rs
```diff
@@ -131,26 +141,55 @@ pub fn TabWebView(tab: Store<Tab>, active_tab_id: Signal<TabId>) -> Element {
...
+                key: "{id}",
+                class: "tab-content",
+                style: visibility,
+                AboutPageView { page, on_navigate }
+            }
+        } else {
+            web-view {
+                key: "{id}",
+                class: "tab-content",
+                style: visibility,
```
nicoburns, line 186: Note that you can do `display: "block"` instead of `style: "display: block"` if you want to.

apps/browser/src/about_pages.rs
```diff
@@ -0,0 +1,152 @@
...
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum AboutPage {
+    NewTab,
+    Settings,
+    History,
+    Bookmarks,
+}
+
+impl AboutPage {
+    pub fn from_url(url: &Url) -> Option<Self> {
```
nicoburns, line 19: Nit: `try_from_url` would probably be a more idiomatic name as this is fallible and returns option.

nicoburns (approved):
I've noted a few quibbles in comments, but overall this looks good :)

PR #413, opened 2026-05-01
feat(browser): browser history and tab favicons

tonybierman:
Adds a browsing history page (`about:history`) with per-entry favicons and elapsed-time labels, and favicon display in the tab strip. Favicon probing/sniffing lives in a new `favicon.rs` module; history data and dedup logic are in `browser_history.rs`.

### DIFF
--- a/apps/browser/Cargo.toml
+++ b/apps/browser/Cargo.toml
@@ -10,13 +10,9 @@ name = "blitz"
 path = "src/main.rs"
 
 [features]
-default = ["vello", "floats", "incremental", "cookies", "cache", "screenshot", "capture", "ico"]
+default = ["vello", "floats", "incremental", "cookies", "cache", "screenshot", "capture"]
 screenshot = ["dep:anyrender", "dep:anyrender_vello_cpu", "dep:blitz-paint", "dep:png", "dep:peniko", "dep:rfd"]
 capture = ["dep:anyrender", "dep:anyrender_serialize", "dep:blitz-paint", "dep:png", "dep:peniko", "dep:rfd"]
-# Enables the `image` crate's ICO decoder so the renderer can actually
-# display .ico favicons (the bulk of real-world favicons). Without this,
-# <img src="…/favicon.ico"> would silently render nothing.
-ico = ["image/ico"]
 vello = ["dioxus-native/vello"]
 hybrid = ["dioxus-native/vello-hybrid"]
 skia = ["dioxus-native/skia"]
@@ -55,7 +51,8 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true, optional = true }
 webbrowser = { workspace = true }
 winit = { workspace = true }
-image = { workspace = true }
+image = { workspace = true, features = ["ico"] }
+usvg = { workspace = true }
 anyrender = { workspace = true, optional = true }
 anyrender_serialize = { workspace = true, optional = true }
 anyrender_vello_cpu = { workspace = true, optional = true }
--- a/apps/browser/Cargo.toml
+++ b/apps/browser/Cargo.toml
@@ -10,9 +10,13 @@ name = "blitz"
 path = "src/main.rs"
 
 [features]
-default = ["vello", "floats", "incremental", "cookies", "cache", "screenshot", "capture"]
+default = ["vello", "floats", "incremental", "cookies", "cache", "screenshot", "capture", "ico"]
 screenshot = ["dep:anyrender", "dep:anyrender_vello_cpu", "dep:blitz-paint", "dep:png", "dep:peniko", "dep:rfd"]
 capture = ["dep:anyrender", "dep:anyrender_serialize", "dep:blitz-paint", "dep:png", "dep:peniko", "dep:rfd"]
+# Enables the `image` crate's ICO decoder so the renderer can actually
+# display .ico favicons (the bulk of real-world favicons). Without this,
+# <img src="…/favicon.ico"> would silently render nothing.
+ico = ["image/ico"]
 vello = ["dioxus-native/vello"]
 hybrid = ["dioxus-native/vello-hybrid"]
 skia = ["dioxus-native/skia"]
--- a/apps/browser/src/browser_history.rs
+++ b/apps/browser/src/browser_history.rs
@@ -105,7 +105,6 @@ pub fn format_elapsed(visited_at: SystemTime, now: SystemTime) -> String {
 }
 
 #[cfg(test)]
-#[allow(clippy::unwrap_used, clippy::indexing_slicing)]
 mod tests {
     use super::*;
 
--- a/apps/browser/src/browser_history.rs
+++ b/apps/browser/src/browser_history.rs
@@ -1,0 +1,141 @@
+use std::collections::VecDeque;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::{Duration, SystemTime};
+
+use blitz_traits::net::Url;
+
+const MAX_HISTORY_ENTRIES: usize = 1000;
+
+const SECONDS_PER_MINUTE: u64 = 60;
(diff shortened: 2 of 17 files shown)


apps/browser/src/favicon.rs
```diff
@@ -0,0 +1,170 @@
...
+        .fetch_async(Request::get(url.clone()))
+        .await
+        .ok()?;
+    if bytes.is_empty() || !looks_like_image(&bytes) {
+        return None;
+    }
+    Some(url.clone())
+}
+
+fn looks_like_image(bytes: &[u8]) -> bool {
```
nicoburns: Could you make the fetching code work by attempting to parse the image. Like so: https://github.com/DioxusLabs/blitz/blob/main/packages/blitz-dom/src/net.rs#L445. That way we won't have to maintain this sniffing code (and we'll have the case where something looks like an image but is actually broken).

`blitz-dom` has this code anyway, so we're not pulling it any extra dependencies. At some point we should probably pull this out into a separate `blitz-media` crate that we could call here. But I think just copy-pasting here would be fine for now (or you could expose it as a public function from `blitz-dom` if you wanted).

apps/browser/src/favicon.rs
```diff
@@ -0,0 +1,170 @@
...
+    })
+}
+
+// Resolved favicon URLs that have been confirmed to fetch as image bytes.
+// Shared across all tabs so navigating around the same site doesn't refetch
+// favicon.ico repeatedly. Only positive results are cached — a transient
+// probe failure (network blip, slow TLS handshake) must not poison the
+// entry for the rest of the process lifetime.
+static FAVICON_CACHE: LazyLock<RwLock<HashMap<Url, Url>>> =
+    LazyLock::new(|| RwLock::new(HashMap::new()));
```
nicoburns, line 15: We do already have a network cache, so not sure this is really required, but it's also fine.

apps/browser/Cargo.toml
```diff
@@ -10,9 +10,13 @@ name = "blitz"
...
 
 [features]
-default = ["vello", "floats", "incremental", "cookies", "cache", "screenshot", "capture"]
+default = ["vello", "floats", "incremental", "cookies", "cache", "screenshot", "capture", "ico"]
 screenshot = ["dep:anyrender", "dep:anyrender_vello_cpu", "dep:blitz-paint", "dep:png", "dep:peniko", "dep:rfd"]
 capture = ["dep:anyrender", "dep:anyrender_serialize", "dep:blitz-paint", "dep:png", "dep:peniko", "dep:rfd"]
+# Enables the `image` crate's ICO decoder so the renderer can actually
+# display .ico favicons (the bulk of real-world favicons). Without this,
+# <img src="…/favicon.ico"> would silently render nothing.
+ico = ["image/ico"]
```
nicoburns: I'd probably make this an unconditional dependency of this crate tbh (you do this by adding a direct dependency on the `image` crate with the ico feature enabled).

apps/browser/src/favicon.rs
```diff
@@ -0,0 +1,170 @@
...
+        let rel = data.attr(local_name!("rel"))?;
+        if !rel
+            .split_ascii_whitespace()
+            .any(|v| v.eq_ignore_ascii_case("icon"))
+        {
+            return None;
+        }
+        data.attr(local_name!("href")).map(|s| s.to_string())
+    })
+}
```
nicoburns: Can we make this a `favicon_url` method on `BaseDocument`?

apps/browser/src/browser_history.rs
```diff
@@ -0,0 +1,141 @@
...
+    COUNTER.fetch_add(1, Ordering::Relaxed)
+}
+
+// Dedupe policy: only consecutive visits to the same URL fold into the head
+// entry; non-consecutive revisits are recorded as new entries. This keeps the
+// list a faithful chronological log rather than a most-recently-visited set,
+// matching the simplest behavior a user can predict at a glance. The id of
+// the existing head entry is preserved on fold so its rendered row is not
+// remounted.
+pub fn record_visit(history: &mut VecDeque<HistoryEntry>, entry: HistoryEntry) {
```
nicoburns: Non-blocking, but I believe if you used `Store<VecDeque<HistoryEntry>>` rather than `Signal<VecDeque<HistoryEntry>>` then I believe these could become methods.

See: https://docs.rs/dioxus-stores/latest/dioxus_stores/attr.store.html#example

nicoburns (changes requested):
This is mostly looking good. However, I have left some review feedback.

nicoburns:
There seems to be an issue with the UI layout when there are enough history entries to cause the page to scroll:

https://github.com/user-attachments/assets/d8e02917-02c8-4456-9618-aa38c1634f40

apps/browser/src/browser_history.rs
```diff
@@ -38,13 +39,38 @@ fn next_history_entry_id() -> HistoryEntryId {
...
 }
 
+#[derive(Default, Store)]
+pub struct BrowsingHistory {
+    pub entries: VecDeque<HistoryEntry>,
+}
+
+#[store(pub)]
+impl<Lens> Store<BrowsingHistory, Lens> {
+    fn record_visit(&self, entry: HistoryEntry)
```
nicoburns, line 49: I believe if you use `&mut self` then you don't need the `where Lens: Writable`

apps/browser/src/capture.rs
```diff
@@ -72,10 +72,24 @@ pub(crate) fn capture_anyrender_scene(doc: &blitz_dom::BaseDocument, path: &Path
     let config = SerializeConfig::new()
```
nicoburns, line 72: The changes in this file are seem unrelated to the rest of the PR and are not needed. Please undo.

apps/browser/src/document_loader.rs
```diff
@@ -74,7 +74,8 @@ impl DocumentLoader {
 
     pub fn reload(&self) {
         let mut reload_generation = self.reload_generation;
-        *reload_generation.write() += 1;
+        let next = (*reload_generation.read()).wrapping_add(1);
+        *reload_generation.write() = next;
     }
```
nicoburns, line 78: No need for wrapping_add. Please undo.

apps/browser/src/fps_overlay.rs
```diff
@@ -41,8 +43,12 @@ impl FpsStats {
...
         }
-        let total: Duration = self.ring[..self.count].iter().copied().sum();
-        let avg = total / self.count as u32;
+        let slice = self.ring.get(..self.count).unwrap_or(&[]);
+        let total: Duration = slice.iter().copied().sum();
+        // count > 0 by the early return above, so division is safe.
+        let avg = total
+            .checked_div(self.count as u32)
+            .unwrap_or(Duration::ZERO);
         let ms = avg.as_secs_f32() * 1000.0;
```
nicoburns, line 46: Move drive-by changes!

apps/browser/src/tab.rs
```diff
@@ -112,17 +114,21 @@ pub fn open_tab(
         favicon_url: None,
     });
 
-    let len = tabs.len();
-    let tab_lens = tabs.get(len - 1).expect("just pushed");
+    // We just pushed; the last element is the tab we want.
+    #[allow(clippy::expect_used)]
```
nicoburns, line 118: Why have all of these popped up, even though that setting isn't in the clippy config anymore?

nicoburns (changes requested):
The favicon, history store changes and layout fix look good.

However, this commit contains a lot of drive-by lint-related changes again. Can we remove those? (and somehow instruct Claude not to keep introducing them?). It's important for maintaintability by humans that the code stays readable, and making 1 liners into 3 or 4 liners does not help with that!

nicoburns:
> this commit contains a lot of drive-by lint-related changes again. Can we remove those?

Realised I could quite easily do this myself using git.

nicoburns (approved):

PR #414, opened 2026-05-03
perf(browser): move favicon probe off the page-load critical path

tonybierman:
Page navigations were blocking on a favicon fetch plus a full image decode before swapping in the new document, adding a round-trip of latency to every load. This moves the probe into a background task spawned after the document is applied, with stale-probe guards on the tab and id-keyed patching of history entries so a late-arriving icon still updates the right row.


tonybierman:
@nicoburns see details in commit.  The performance hit was sluggish enough to merit a quick PR to fix it IMHO.

nicoburns (approved):

nicoburns:
Thanks!

PR #415, opened 2026-05-03
feat(browser): urlbar autocomplete and on-disk history

tonybierman:
Adds urlbar autocomplete (history matches + a "Search with DuckDuckGo" row, keyboard-driven) backed by a new sqlite history store that hydrates the in-memory `BrowsingHistory` on startup.


apps/browser/src/history_store.rs
```diff
@@ -0,0 +1,549 @@
...
+    fn in_memory_fallback() -> Connection {
+        Connection::open_in_memory().expect("sqlite in-memory open must not fail")
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Mobile stub (Android and iOS)
+// ---------------------------------------------------------------------------
+#[cfg(any(target_os = "android", target_os = "ios"))]
+mod mobile {
```
nicoburns: The only limitation on mobile is that we don't have easy access to directory to store the database in (this can also be made to work, we just need to workout how). You should be able to use the in-memory database just fine on mobile. Might make sense to do that rather than this stub?

apps/browser/src/history_store.rs
```diff
@@ -0,0 +1,549 @@
...
+            }
+        }
+    }
+
+    pub(super) fn load_recent_inner(conn: &Connection, limit: usize) -> Vec<HistoryEntry> {
+        let mut stmt = match conn.prepare(
+            "SELECT url, title, favicon_url, visited_at \
+             FROM history_entries \
+             ORDER BY visited_at DESC \
+             LIMIT ?1",
```
nicoburns: A note that `\` at the end of the line of a multiline string is not necessary in Rust (although it is valid). Without the `\`, you will end up with a newline in the string (which is valid SQL). With the `\` there will be no newline.

apps/browser/assets/browser.css
```diff
@@ -302,3 +302,47 @@ body,
   color: #333;
   z-index: 10;
 }
+
```
nicoburns, line 305: Just a note that the suggestions are looking great!

apps/browser/src/toolbar.rs
```diff
@@ -259,61 +335,81 @@ pub fn Toolbar(
...
+                    oninput: move |evt| {
+                        *url_input_value.write() = evt.value();
+                        selected_suggestion.set(None);
+                    },
+                }
+                if is_focused() && !suggestions.read().is_empty() {
+                    UrlSuggestions {
+                        suggestions,
+                        selected_idx: selected_suggestion,
+                        on_hover: move |idx| selected_suggestion.set(idx),
```
nicoburns: I don't think the selected_suggestion should be set on hover. If you test how chrome/firefox suggestions, then hovering a suggestion does highlight it, but only visually (this could just be a CSS `:hover` style), and the item selected by the keyboard remains selected.

I think the clicking action could be implemented directly rather than setting the selected_selection (or the selected_selection could be set as part of the click handler.

apps/browser/src/url_suggestions.rs
```diff
@@ -0,0 +1,314 @@
...
+/// rows (deduped by URL, most-recent first) followed by exactly one Search row.
+pub fn build_suggestions(query: &str, recent: &VecDeque<HistoryEntry>) -> Vec<Suggestion> {
+    if query.is_empty() {
+        return vec![];
+    }
+
+    // Lowercase the needle once. Matching against entries uses
+    // `contains_ignore_ascii_case` so we don't allocate per-entry — at up to
+    // ~1000 history rows, three String allocs per entry per keystroke adds up.
+    let q_lower = query.to_ascii_lowercase();
```
nicoburns: We should probably eventually switch to a fuzzy searching library like [`nucleo`](https://github.com/helix-editor/nucleo), but this is fine for now.

apps/browser/src/history_store.rs
```diff
@@ -0,0 +1,549 @@
...
+               visited_at  INTEGER NOT NULL
+             );
+             CREATE INDEX idx_history_visited_at ON history_entries(visited_at DESC);
+             CREATE INDEX idx_history_url        ON history_entries(url);
+             PRAGMA user_version = 1;
+             COMMIT;",
+        )
+    }
+
+    fn open_connection() -> Connection {
```
nicoburns: It would be possible to make this less verbose by using the `?` operator along with the `anyhow` crate. Something like:

```rust
fn open_file_connection() -> anyhow::Result<Connection> {
    let path = ProjectDirs::from("com", "DioxusLabs", "Blitz")
        .map(|p| p.data_dir().join("history.sqlite3"));

    let db_path = path.ok_or(|| anyhow!("ProjectDirs::from returned None."))?;
    let parent = db_path.parent().ok_or(|| anyhow!("No parent path"))?;
    std::fs::create_dir_all(parent)
        .with_context(|| format!("could not create data dir {}", parent.display()))?;

    Connection::open(db_path)
        .with_context(|| format!("could not open {}", db_path.display()))?;
}

fn open_connection() -> Connection {
    match open_file_connection() {
        Ok(conn) => conn,
        Err(err) => {
            tracing::warn!("history_store: {:#} falling back to in-memory", err);
            in_memory_fallback()
        }
    }
}
```

I guess using `Result<Connection, String>` would also work.

apps/browser/src/history_store.rs
```diff
@@ -0,0 +1,549 @@
...
+        if let Err(e) = conn.execute(
+            "UPDATE history_entries SET favicon_url = ?1 \
+             WHERE url = ?2 AND favicon_url IS NULL",
+            params![favicon_url.as_str(), page_url.as_str()],
+        ) {
+            tracing::warn!("history_store: set_favicon_for_url failed: {e}");
+        }
+    }
+
+    pub(super) fn bootstrap_schema(conn: &Connection) -> rusqlite::Result<()> {
```
nicoburns: This is fine, but we may want to use this crate (https://docs.rs/rusqlite_migration) for a slightly nicer API at some point.

apps/browser/src/history_store.rs
```diff
@@ -0,0 +1,549 @@
+//! On-disk browsing-history store.
```
nicoburns: Also doesn't need to happen in this PR, but I'm thinking that it might make sense to make the persistence layer into it's own crate?

apps/browser/src/url_suggestions.rs
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+
+    let mut result = history_rows;
+    result.push(Suggestion {
+        kind: SuggestionKind::Search,
+        display_title: format!("Search with {SEARCH_ENGINE_NAME}: {query}"),
+        display_subtitle: String::new(),
+    });
+    result
```
nicoburns: Would it be possible to get an additional suggestion row (at the top) which has the action of "whatever submitting the url bar without selecting a suggestion would do". Currently it isn't possible to get back to that behaviour if you start selecting a suggestion.

nicoburns (commented):
A few suggestions. Nothing blocking, but a few that would clean up the code. Action any that you want to, and leave any that you don't and then ping me for re-review?

tonybierman:
The Android failure is not a code issue — it's a transient CI infrastructure failure.

tonybierman:
@nicoburns ready for re-review.  significant changes.

nicoburns (approved):
I think we'd be a lot better off using the `nucleo` crate rather than `nucleo-matcher`. It  does the hard part of computing suggestions in a background thread for us. And even has an API that allows us to retrieve results incrementally as they're computed. But this is fine for now :)

PR #416, opened 2026-05-04
perf(browser): move urlbar fuzzy matching to nucleo worker

tonybierman:
Replaces the synchronous `nucleo-matcher` rescore in the urlbar with a tokio worker driving the higher-level `nucleo` crate, so typing no longer rescans full history on the UI thread.


tonybierman:
@nicoburns request review

nicoburns (approved):
Not sure that the tests are adding much value here, but the actual code looks good, and we can always delete the tests later if we want to.

PR #418, opened 2026-05-05
syntax-highlighting codeblocks for readme example

tonybierman:
added syntax highlighting to readme example as per roadmap


tonybierman:
@nicoburns request review. 

I don’t like the way this code smells.  If I were doing it from greenfield, I’d approach it differently.  But I didn’t think you’d want wholesale change, so I shoe-horned syntax highlighting in with the existing stuff.  I'm sure this wants to be changed a bit, but I wanted your guidance.  It’s on the roadmap, but maybe not a priority for you to spend time on right now.  If so, NBD; I’ll discard this and work on something else.

(I think even the name of the app is wrong; it’s called readme, but the code can load any md file)

PR #420, opened 2026-05-05
fix windows icon for browser

Klemen2:
The new cli now takes the windows icon from bundle icon
Also matches https://github.com/DioxusLabs/dioxus/pull/3753 to display window icon


nicoburns (approved):
Thanks! I have no way of testing this, but I assume you have, and will merge on trust assuming CI passes :)

PR #425, opened 2026-05-08
New Custom Widget API

nicoburns:
Replaces the `CustomPaintSource` API with a new Custom Widget API based based on the `Widget` trait. See `custom_widget` and `wgpu_texture` examples for details.


jkelleyrtp:
FWIW I think having the ability to use WGPU into canvas elements is still a nice API if we can support it!

nicoburns:
@jkelleyrtp Supporting (a slight variant on) the old `CustomPaintSource` trait and `<canvas>` as the blitz-dom API would be trivial (`CustomPaintSource` functionality is a subset of `Widget` functionality, so internally we'd just wrap it in a `Widget`). But a `use_wgpu` hook would now be doing barely anything as we no longer need access to the renderer from the Dioxus side (that's now all handled on the Blitz side).

It would just be something like:

```rust
fn use_wgpu(custom_paint_source: impl CustomPaintSource) -> CanvasSourceAttr {
     use_hook(|| CustomPaintSourceAttr::new(custom_paint_source);
}
```

What I would like to add `blitz-wgpu-widget` crate with a `WGPUWidget` trait which wraps the more general `Widget` trait and handles the downcasting to `wgpu` types for you.

PR #426, opened 2026-05-08
Add wasm32 support (async renderer resume)

tonybierman:
Adds `View::resume_async` and a `BlitzWasmApplication` driver so blitz can run on `wasm32-unknown-unknown` without `pollster::block_on` deadlocking the JS event loop on wgpu init; `examples/wasm_hello` is a minimal Trunk-served demo. Draft pending DioxusLabs/anyrender#59 — the `[patch.crates-io]` git deps will be swapped for crates.io versions once that PR merges and the new `anyrender` crates are released.


tonybierman:
@nicoburns this is setup for you to run, not to be merged.  There's a bundled font in the example, it patches AnyRender with my fork, and more.  I'm sure you'll find a lot you don't like, but it does run.  I've been running with `trunk serve`.  It should run for you out of the box - it does for me.

nicoburns (changes requested):
@tonybierman This is a good start! The setup running in WASM is really helpful!

I think there might be a simpler approach, which would allow us to continue using the regular `BlitzApplication` on WASM (and would mostly unify the implementations between WASM and native):

- Rather than creating an `async fn` to do the asynchronous parts, make `resume` take a callback, spawn a future internally,. and then call the callback when the future resolves (or in the native case, simply at the end of the resume function).
- The `WindowRenderer` should internally own a channel, and then send the created surface back through the channel (to itself) when the async future resolves.
- When the callback is called, have `Window`/`View` send a new custom blitz event (add to same enum as `Poll`, etc). Let's same `ResumeCompleted` or `ResumeReady` or something.
- In the handler for this new event, call a new method on `WindowRenderer` which checks the channel, and then completes the initialization with the surface that should be in there.

What do you think?

tonybierman:
@nicoburns request re-review

image

nicoburns:
@tonybierman Your latest change doesn't compile because you need to change the filename in the import to `.woff`:

```rust
const DEJAVU_SANS: &[u8] = include_bytes!("../assets/DejaVuSans.ttf");
```

Relatedly: we should probably add a CI job for the WASM target!

tonybierman:
@nicoburns I should have caught the ttf. I fixed it and added a CI job for the wasm examples.

tonybierman:
@nicoburns i made a minor button fix in one of the examples on top of your force push

PR #427, opened 2026-05-08
Cross origin iframe support

gterzian:
- layout fixes so iframes actually show up
- rendering changes to support cross origin iframes, and integrate with https://github.com/DioxusLabs/anyrender/pull/60


gterzian:
https://github.com/DioxusLabs/anyrender/pull/60

PR #428, opened 2026-05-10
Upgrade to WGPU v29 + AnyRender 0.10 (+ font embolden)

nicoburns:


tonybierman:
@nicoburns want me to fix the wasm build?

tonybierman:
@nicoburns The **Build wasm examples** failure is caused by a duplicate `anyrender` in the dependency graph. `examples/wasm_hello` and `examples/wasm_7guis` declare their own `[workspace]`, so they don't inherit the root's git overrides. Their crates.io `anyrender` and the git `anyrender` pulled in via `blitz-shell` are treated as two separate crates, breaking the trait bounds.

The immediate fix is a `[patch.crates-io]` block in each example, but that has to be manually synced every time a workspace-level dep changes — as this PR demonstrates.

A cleaner structural fix: add the wasm examples to the root workspace as non-default members.

```toml
# root Cargo.toml
[workspace]
members = [
    # ... existing members ...
    "examples/wasm_hello",
    "examples/wasm_7guis",
]
default-members = [
    # ... existing default members, excluding the wasm examples ...
]
```

With this, the examples inherit `[workspace.dependencies]` and `[patch.crates-io]` automatically. Native `cargo build` / `cargo check` respect `default-members` and skip them; the dedicated wasm CI job already targets `wasm32-unknown-unknown` explicitly so nothing changes there. The standalone `[workspace]` declarations in the example `Cargo.toml` files would be removed.

nicoburns:
@tonybierman I think it's just that I didn't upgrade to matching anyrender versions there. Was there a reason why the wasm examples aren't in the workspace? Or could we look at moving them into the workspace (where they'd pick up the workspace versions by default)?

tonybierman:
@nicoburns I can add the wasm examples to the root workspace as non-default members.  See my detailed comment above.

nicoburns:
Non-default members makes sense for wasm examples I think

PR #431, opened 2026-05-12
WASM canvas resize: clear winit inline styles and debounce surface reconfigure

tonybierman:
Make Blitz reflow with its host `<canvas>`: clear winit's inline size override,
and debounce the wasm32 surface reconfigure to avoid mid-drag flicker.


tonybierman:
@nicoburns request review

nicoburns:
#432 was merged instead

PR #432, opened 2026-05-12
WASM canvas resize: skip with_surface_size on wasm (alt to #431)

tonybierman:
Alternative to #431 implementing @nicoburns's hypothesis: drop `.with_surface_size(...)` on wasm so winit-web never writes inline `canvas.style.width/height`, letting host CSS size the canvas directly — no style-stripping needed. Same wgpu `surface.configure` debounce as #431 (with timer coalescing to avoid closure leaks during drag), plus a fallback that seeds the initial size from the canvas's CSS layout when winit-web reports 0×0 at init.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -954,6 +954,8 @@ dependencies = [
  "keyboard-types 0.7.0",
  "rfd",
  "tracing",
+ "wasm-bindgen",
+ "web-sys",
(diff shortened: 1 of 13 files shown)


tonybierman:
@nicoburns request review

nicoburns (approved):

PR #434, opened 2026-05-12
Make dioxus_native::launch_cfg work on WASM

tonybierman:
Adds `Config::with_font_ctx` and seeds a `with_append(true)` wasm canvas default so `dioxus_native::launch_cfg` compiles and runs on `wasm32` (with the tokio-backed `blitz_net::Provider` gated off there). The seven_guis and todomvc examples collapse their hand-rolled wasm entry points to a single `launch_cfg` call.

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -343,7 +343,10 @@ impl BaseDocument {
                     source_cache: SourceCache::new_shared(),
                     collection: Collection::new(CollectionOptions {
                         shared: false,
-                        system_fonts: true,
+                        system_fonts: cfg!(all(
+                            feature = "system_fonts",
+                            not(target_arch = "wasm32")
+                        )),
                     }),
                 };
                 font_ctx
--- a/packages/blitz-dom/src/lib.rs
+++ b/packages/blitz-dom/src/lib.rs
@@ -90,3 +90,36 @@ pub type SelectorList = selectors::SelectorList<style::selector_parser::Selector
 pub use events::{EventDriver, EventHandler, NoopEventHandler};
 pub use html::{DummyHtmlParserProvider, HtmlParserProvider};
 pub use util::{Point, decode_font_bytes};
+
+/// Convenience builder for the one-font case: produces a [`FontContext`] with
+/// system-font discovery disabled and the supplied font registered as the
+/// fallback for every generic family. The standard setup for WASM, where
+/// browsers don't expose system fonts. WOFF/WOFF2 inputs are decoded
+/// automatically.
+pub fn build_single_font_ctx(font_data: &[u8]) -> FontContext {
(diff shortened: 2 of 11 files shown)


tonybierman:
@nicoburns request review

packages/dioxus-native/src/lib.rs
```diff
@@ -107,6 +107,9 @@ pub fn launch_cfg_with_props<P: Clone + 'static, M: 'static>(
     contexts: Vec<Box<dyn Fn() -> Box<dyn Any> + Send + Sync>>,
     configs: Vec<Box<dyn Any>>,
 ) {
+    #[cfg(all(target_arch = "wasm32", feature = "panic-hook"))]
+    console_error_panic_hook::set_once();
+
```
nicoburns: I think this should remain in user code. Users may want some other panic hook. And Dioxus/Blitz will function without this.

packages/blitz-dom/src/document.rs
```diff
@@ -343,7 +343,7 @@ impl BaseDocument {
                     source_cache: SourceCache::new_shared(),
                     collection: Collection::new(CollectionOptions {
                         shared: false,
-                        system_fonts: true,
+                        system_fonts: cfg!(not(target_arch = "wasm32")),
```
nicoburns: This should also be gated on the `system_fonts` feature being enabled (pre-existing bug that I noticed while reviewing this).

```rust
cfg!(all(feature = "system_fonts", not(target_arch = "wasm32")))
```

packages/blitz-dom/src/lib.rs
```diff
@@ -90,3 +90,30 @@ pub type SelectorList = selectors::SelectorList<style::selector_parser::Selector
 pub use events::{EventDriver, EventHandler, NoopEventHandler};
 pub use html::{DummyHtmlParserProvider, HtmlParserProvider};
 pub use util::{Point, decode_font_bytes};
+
+/// Build a [`FontContext`] seeded with one font, used as the fallback for every
+/// generic family. System font discovery is disabled, so all text renders with
+/// the supplied font — required on WASM, where browsers don't expose system
+/// fonts. Run WOFF/WOFF2 inputs through [`decode_font_bytes`] first.
+pub fn build_font_ctx(font_data: impl Into<parley::fontique::Blob<u8>>) -> FontContext {
```
nicoburns: Could we call `decode_font_bytes` internally in this function so users don't have to call it?

nicoburns: Perhaps this ought to have a more specific name (`build_single_font_ctx` or something?) It's a bit of a hack and not a general-purpose font context builder...

nicoburns (changes requested):
Looking good - this removes a lot of boilerplate! A few changes requested though.

tonybierman:
@nicoburns request review.

nicoburns (approved):

Issue #430, opened 2026-05-11
Resolving HtmlDocument in many different threads panics


ealmloff:
While testing https://github.com/DioxusLabs/dioxus/pull/5323, I ran into a panic when you create documents in multiple threads at once. From the backtrace it looks like this may be an upstream issue.

If you run this code ([Reproduction repo](https://github.com/ealmloff/blitz-stylo-race))

```rust
use blitz_dom::{Document as _, DocumentConfig};
use blitz_html::HtmlDocument;
use std::thread;

fn main() {
    let html = "<div></div>".repeat(64);
    let handles: Vec<_> = (0..16)
        .map(|_| {
            let html = html.clone();
            thread::spawn(move || {
                for _ in 0..1000 {
                    HtmlDocument::from_html(&html, DocumentConfig::default())
                        .inner_mut()
                        .resolve(0.0);
                }
            })
        })
        .collect();

    for h in handles {
        h.join().expect("a worker thread panicked");
    }
}
```

It panics with this backtrace:
```
thread 'StyleThread#5' (15302138) panicked at /Users/evanalmloff/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/atomic_refcell-0.1.14/src/lib.rs:158:23:
already mutably borrowed
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/panicking.rs:80:14
   2: core::panicking::panic_display
             at /Users/evanalmloff/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/panicking.rs:259:5
   3: atomic_refcell::AtomicRefCell<T>::borrow_mut
             at /Users/evanalmloff/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/atomic_refcell-0.1.14/src/lib.rs:158:23
   4: style::sharing::StyleSharingCache<E>::new::{{closure}}
             at /Users/evanalmloff/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/stylo-0.17.0/sharing/mod.rs:619:50
   5: std::thread::local::LocalKey<T>::try_with
[...]

PR #437, opened 2026-05-12
Add StyleThreading::Sequential opt-out for concurrent document resolves

tonybierman:
Adds a `StyleThreading::Sequential` opt-in on `DocumentConfig` for callers that need to resolve documents from multiple threads concurrently (#430); the gate sits at the `traverse_dom` call site in `packages/blitz-dom/src/stylo.rs` rather than via `layout.threads`, because the pref is read once when `STYLE_THREAD_POOL` is first forced and would silently no-op for any document constructed after the pool exists.

Default stays `Parallel`; verified locally against the issue repro across all-`Sequential`, all-`Parallel` (panics, as a negative control), and pool-already-initialized mixed-mode scenarios.

### DIFF
--- a/apps/browser/src/document_loader.rs
+++ b/apps/browser/src/document_loader.rs
@@ -57,6 +57,7 @@ pub fn make_doc_config(
         html_parser_provider: Some(Arc::new(HtmlProvider)),
         font_ctx: Some(font_ctx),
         media_type: None,
+        ..Default::default()
     }
 }
 
--- a/apps/browser/src/toolbar.rs
+++ b/apps/browser/src/toolbar.rs
@@ -199,6 +199,7 @@ pub fn Toolbar(
             html_parser_provider: Some(Arc::new(HtmlProvider)),
             font_ctx: Some(tab.loader_rc().font_ctx.clone()),
             media_type: None,
+            ..Default::default()
         };
         let mut document = HtmlDocument::from_html(view_source_html, config).into_inner();
         if let Some(parent_id) = document.get_element_by_id("source") {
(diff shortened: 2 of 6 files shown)


nicoburns (approved):
This seems like the right approach. Much nicer that it's per-document :). Approved subject to CI being fixed.

tonybierman:
@nicoburns I closed it before I saw your comment.  I can continue down this route if you think it is the right approach.  i was afraid it would turn into a rabbit hole.  I'll fix the CI and we can take a look.

nicoburns:
I think this is fine. Breaking change is a non-issue as there are already much bigger breaking changes in store for anyone upgrading to the next version of Blitz. There's an argument for different places to pass it, but I think this is a decent option. Probably they will be something that people set on an application-global level in practice.

I'd basically be keen to land *something* that solves this issue to unblock people. And we can figure out the optimal API later.

tonybierman:
Reopening with apps/browser fixup. The previous CI red was entirely downstream of two exhaustive `DocumentConfig` struct literals in apps/browser that hadn't picked up the new `style_threading` field; appending `..Default::default()` resolves them. Verified locally: `cargo check --workspace`, `cargo clippy --workspace -- -D warnings`, `cargo test --workspace`, and `cargo fmt --all -- --check` all clean on Rust 1.92.0 (workspace MSRV).

The breaking-change concern on `DocumentConfig` (any new non-`Option` field breaks external exhaustive struct-literal callers) still applies — happy to switch to a setter-on-`BaseDocument` shape if maintainers prefer that for the public API.

nicoburns:
At some point we should probably make `DocumentConfig` a builder struct to avoid this issue, but for the purposes of this PR I'm quite happy for this to just be a breaking change.

Issue #435, opened 2026-05-12
Images loading inconsistently on wikipedia.org


nicoburns:
If I repeatedly load https://en.wikipedia.org/wiki/Main_Page then the images only load sometimes. And sometimes some images load and others dont. This is very odd, and didn't used to happen.

### Video

https://github.com/user-attachments/assets/5fb44060-cc5d-41a3-9497-90c29217834d

tonybierman:
@nicoburns the issue is `HTTP 429 Too Many Requests`. When I curl 13 URLs in parallel from the same IP with the same UA, Wikimedia returns 429s with a 1964-byte error page — exactly matching what blitz-net is  seeing. Sequential curl works fine (200 OK). It's pure rate limiting, not UA / TLS / cookies / cache.

The deeper bug in blitz-net: it never checks the HTTP status code. In `packages/blitz-net/src/lib.rs::fetch_inner:`

let response = req.await.send().await?;
(response.url().to_string(), response.bytes().await?)

A 429 (or 4xx/5xx of any kind) returns the response body as if it were a successful fetch — that's why every failure logs INFO blitz_net: Success fetching url="...". The error page bytes then fall into the image decoder, which fails with a generic "could not parse".

tonybierman:
@nicoburns While in the code fixing this issue, these additional issues were identified.  I think I should fix them.

[BLITZ_TODO.md](https://github.com/user-attachments/files/27662196/BLITZ_TODO.md)

PR #438, opened 2026-05-12
fix: key image cache on requested URL, not post-redirect URL (refs #435)

tonybierman:
**Thesis:** PR #331's URL-keyed image cache uses the pre-fetch URL on the request side but the post-redirect URL (`reqwest::Response::url()`) on the response side, so any redirected image fetch silently drops its bytes — the likely cause of #435's intermittent image loads on wikipedia.org, where on-disk HTTP-cache hits dodge the redirect and work while fresh fetches that follow a redirect do not. This PR threads the original requested URL through `ResourceLoadResponse` and keys both the image cache and `pending_images` on it; `resolved_url` is dropped from the response struct since nothing else consumed it.

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -843,6 +843,7 @@ impl BaseDocument {
                                 self.tx.clone(),
                                 self.id,
                                 Some(node_id),
+                                resolved_href.to_string(),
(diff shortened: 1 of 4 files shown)


tonybierman:
@nicoburns here's a thesis

PR #439, opened 2026-05-12
Fix intermittent image loads on rate-limited CDNs (#435)

### LINKED ISSUE #435: Images loading inconsistently on wikipedia.org

tonybierman:
Caps per-host concurrency at 6 with 429/503 retry-and-backoff (honoring `Retry-After`), updates the User-Agent per Wikimedia policy, and surfaces image decode/HTTP errors that were previously swallowed — together fixing intermittent image load failures on rate-limited CDNs. The browser app also gains an optional CLI URL argument for quick reproduction.

Fixes #435.


nicoburns (changes requested):
This seems to go a long way towards fixing the issue for me. If I repeatedly refresh the page I can still hit the rate limit, but if I'm being sensible it is now working well.

There is however one issue: if I try to close the browser while there is an active "retry" pending/in-progress then it hangs. Separately, I looked at what other browsers do when they hit a 429/503, and it seems that they don't automatically retry. They just treat it as an error (and allow the user to retry by refreshing the page).

I therefore suggest that we keep the rate-limiting code here, but remove the retry code and then:

- Setup http2
- See if there's more we do to cache locally (why are already-fetched images not being served from the local http cache?)
- Wire through an error path for network requests so that `blitz-dom` is informed of failed requests.

tonybierman:
@nicoburns I like that plan.  I’m assuming the three bullet points are follow-on, individual PRs, correct? I’ll proceed to remove the retry code, fix the hang on retry when closing, and resubmit this PR.

nicoburns:
Correct :)

tonybierman:
@nicoburns it made sense to me to fix caching while I was handling mid-flight browser refreshes.  I hope you don’t mind.  Request review.

HTTP/2 should be a quick turnaround; I did most the work last night. Error path wiring will follow that.

nicoburns:
@tonybierman  Hmm... I'm not sure that we want to share the image cache between documents. That opens up a whole can of worms around continually growing memory usage. We have disk-based http-level caching in the net provider that ought to be good enough for "new page load" scenarios without balooning memory usage.

tonybierman:
@nicoburns ok, I'll revert the SharedImageCache half of 3193e260 (keep the abort wiring),  and dig into the disk-cache behavior as a follow-up PR.

nicoburns (approved):
Not sure about the UserAgent change. But probably we ought to make this more configurable anyway.

PR #440, opened 2026-05-13
blitz-net: restore Firefox UA to fix DuckDuckGo redirect links

tonybierman:
Bug with DuckDuckGo links. Reverted UA to Firefox.


tonybierman:
@nicoburns request review

PR #441, opened 2026-05-14
Add blitz-net Provider integration tests

tonybierman:
Adds a workspace-member test crate at `tests/blitz-net/` exercising the public `Provider` API directly with wiremock — scheme dispatch, HTTP status mapping, body encoding, abort signals, per-host concurrency, redirects, and feature-gated paths (cookies, cache, multipart). WPT covers rendering conformance but its runner bypasses `blitz-net` entirely (custom NetProvider, pixel-diff assertions), so this complements WPT for structural API regression rather than overlapping it.


tonybierman:
@nicoburns Noted your comment about not getting to this until after RustWeek; I can rebase this then as needed.

Issue #444, opened 2026-05-18
Incorrect scroll height calculation clamps overflow scrolling prematurely


notdanilo:
#### The Bug
Currently, scrollable containers (elements with overflow-y set to scroll or auto, as well as html/body tags with visible) do not always compute their full scrollable range correctly.

Relying solely on final_layout.scroll_height() can underestimate the total height of nested children, leading to early scroll clamping. Users are unable to scroll all the way down to see the bottom-most contents of a container.

#### Steps to Reproduce
1. Create a container with a fixed height and overflow-y: auto.
2. Populate it with absolute or relative layout children that extend beyond the container's height.
3. Attempt to scroll to the bottom. The scroll area is clamped early, clipping the bottom-most children.

```html
<!-- A scrollable container with a fixed height and overflow-y: auto -->
<div style="height: 200px; overflow-y: auto; border: 2px solid #333; padding: 10px;">
    <!-- Nested layout children whose heights clearly exceed the 200px boundary -->
    <div style="height: 150px; background-color: #ff6b6b; margin-bottom: 15px;">
        Child Element 1 (150px)
    </div>
    <div style="height: 150px; background-color: #4dadf7; margin-bottom: 15px;">
        Child Element 2 (150px)
    </div>
    <div style="height: 100px; background-color: #51cf66;">
        Child Element 3 (100px) - Clipped and unreachable via scroll
    </div>
</div>
```

#### Expected Behavior
The scrollable height of a container should span the full bottom coordinate of all layout children plus any padding and border bounds.

yinnho:
Same symptom class here, from the worse end: my engine (also taffy-based) had no scroll range at all — scrollHeight/scrollWidth just returned the element's own box size, so scrollHeight - clientHeight was always 0 and nothing (my own wheel-scroll helper included) could ever detect scrollability.

Fixed it this week: the scroll extent is now computed after layout as the union of the element's own rect with every laid-out descendant rect, relative to the element origin. For html/body I clamp the result up to the viewport, so a short page reports documentElement.scrollHeight == innerHeight (Chrome's "scrolling area is never smaller than the window") while clientWidth/clientHeight keep the viewport contract — that keeps scrollHeight - clientHeight working as an overflow probe. One deliberate non-feature: scrollTop is not clamped to the new range, because synthetic clamping deadlocked scroll-driven lazy loaders on my side (content that only mounts as you scroll never gets a chance to).

[...]

PR #445, opened 2026-05-18
Fix container scroll bounds by factoring child layout heights

notdanilo:
#### Proposed Changes
This PR fixes the scroll clamping issue #444 by computing a corrected scrollable height for scrollable containers based on their layout children's actual vertical bounds:

True Scrollable Area Calculation: Iterates through the container's layout children to find the bottom-most point (location.y + size.height) and factors in padding and border height. The maximum of this calculation and the default layout scroll height is used.

#### Impact
Ensures scrollable containers allow users to scroll fully to the bottom of all content.

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -1537,15 +1537,14 @@ impl BaseDocument {
         y: f64,
         mut dispatch_event: F,
     ) -> bool {
-        let Some(node) = self.nodes.get_mut(node_id) else {
+        // Pre-compute scroll capabilities and corrected scroll height using shared references (before mutable borrow)
+        let Some(node) = self.nodes.get(node_id) else {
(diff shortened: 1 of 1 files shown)


nicoburns:
Hmm.. this seems to be compensating for the fact that Taffy's content size computation is wrong. I think we'd be better fixing it there. In your mind, is this still needed if we merge https://github.com/DioxusLabs/taffy/pull/955 (or https://github.com/DioxusLabs/taffy/pull/956)?

(separately, we may need to move content size / scroll overflow calculation out of Taffy and into Blitz so that we can account for transforms. But that should be a wider change that happens ahead of time and incrementally rather than being just-in-time computation on every mousemove event)

nicoburns:
I'm not 100% sure, but I think this may now have been fixed by work on transforms.

PR #446, opened 2026-05-24
Fix 2d transform hit test & rendering children

Klemen2:
This also needs a redraw call somewhere since live updating the transform doesn't update it in real-time, but I cannot figure out where

https://github.com/user-attachments/assets/966a6725-1bf7-4107-82a5-c18332b2dec3

### DIFF
--- a/packages/blitz-dom/src/layout/mod.rs
+++ b/packages/blitz-dom/src/layout/mod.rs
@@ -450,9 +450,7 @@ impl RoundTree for BaseDocument {
     }
 
     fn set_final_layout(&mut self, node_id: NodeId, layout: &Layout) {
-        let node = self.node_from_id_mut(node_id);
-        node.final_layout = *layout;
-        node.set_transform();
+        self.node_from_id_mut(node_id).final_layout = *layout;
     }
 }
 
--- a/packages/blitz-dom/src/layout/mod.rs
+++ b/packages/blitz-dom/src/layout/mod.rs
@@ -450,7 +450,9 @@ impl RoundTree for BaseDocument {
     }
 
     fn set_final_layout(&mut self, node_id: NodeId, layout: &Layout) {
-        self.node_from_id_mut(node_id).final_layout = *layout;
+        let node = self.node_from_id_mut(node_id);
+        node.final_layout = *layout;
+        node.set_transform();
     }
 }
 
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -154,6 +154,14 @@ impl BaseDocument {
                 overflow = overflow.union(child_rect_in_self);
             }
         }
+        if let Some(before) = self.nodes[node_id].before {
+            let child_rect_in_self = self.resolve_transforms(before);
+            overflow = overflow.union(child_rect_in_self);
+        }
+        if let Some(after) = self.nodes[node_id].after {
+            let child_rect_in_self = self.resolve_transforms(after);
+            overflow = overflow.union(child_rect_in_self);
+        }
 
         self.nodes[node_id].scrollable_overflow = overflow;
         *self.nodes[node_id].layout_children.get_mut() = layout_children;
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -129,12 +129,24 @@ impl BaseDocument {
             return;
         }
 
+        if !self.nodes[node_id]
+            .damage()
+            .map(|d| d.contains(RestyleDamage::RECALCULATE_OVERFLOW))
+            .unwrap_or(false)
+        {
+            return;
+        }
+
         self.nodes[node_id].set_transform();
 
-        let children = self.nodes[node_id].children.clone();
-        for child_id in children {
-            self.resolve_transforms(child_id);
+        let layout_children = std::mem::take(self.nodes[node_id].layout_children.get_mut());
+        if let Some(lc) = layout_children.as_ref() {
+            for child_id in lc {
+                self.resolve_transforms(*child_id);
+            }
         }
+
+        *self.nodes[node_id].layout_children.get_mut() = layout_children;
 
         if let Some(before) = self.nodes[node_id].before {
             self.resolve_transforms(before);
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -147,13 +147,6 @@ impl BaseDocument {
         }
 
         *self.nodes[node_id].layout_children.get_mut() = layout_children;
-
-        if let Some(before) = self.nodes[node_id].before {
-            self.resolve_transforms(before);
-        }
-        if let Some(after) = self.nodes[node_id].after {
-            self.resolve_transforms(after);
-        }
     }
 
     pub fn resolve_scroll_animation(&mut self) {
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -81,6 +81,9 @@ impl BaseDocument {
         self.resolve_layout();
         timer.record_time("layout");
 
+        self.resolve_transforms(root_node_id);
+        timer.record_time("transform");
+
         // Clear all damage and dirty flags
         #[cfg(feature = "incremental")]
         {
@@ -119,6 +122,26 @@ impl BaseDocument {
         timer.record_time("subdocs");
 
         timer.print_times(&format!("Resolve({}): ", self.id()));
+    }
+
+    fn resolve_transforms(&mut self, node_id: usize) {
+        if !self.nodes.contains(node_id) {
+            return;
+        }
+
+        self.nodes[node_id].set_transform();
+
+        let children = self.nodes[node_id].children.clone();
+        for child_id in children {
+            self.resolve_transforms(child_id);
+        }
+
+        if let Some(before) = self.nodes[node_id].before {
+            self.resolve_transforms(before);
+        }
+        if let Some(after) = self.nodes[node_id].after {
+            self.resolve_transforms(after);
+        }
     }
 
     pub fn resolve_scroll_animation(&mut self) {
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -131,7 +131,7 @@ impl BaseDocument {
 
         if !self.nodes[node_id]
             .damage()
-            .map(|d| d.contains(RestyleDamage::RECALCULATE_OVERFLOW))
+            .map(|d| d.contains(style::selector_parser::RestyleDamage::RECALCULATE_OVERFLOW))
             .unwrap_or(false)
         {
             return;
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -6,6 +6,7 @@ use std::{
 };
 
 use debug_timer::debug_timer;
+use kurbo::{Affine, Rect};
 use parley::LayoutContext;
 use selectors::Element as _;
 use style::dom::TDocument;
@@ -124,9 +125,9 @@ impl BaseDocument {
         timer.print_times(&format!("Resolve({}): ", self.id()));
     }
 
-    fn resolve_transforms(&mut self, node_id: usize) {
+    fn resolve_transforms(&mut self, node_id: usize) -> Rect {
         if !self.nodes.contains(node_id) {
-            return;
+            return Rect::ZERO;
         }
 
         if !self.nodes[node_id]
@@ -134,19 +135,39 @@ impl BaseDocument {
             .map(|d| d.contains(style::selector_parser::RestyleDamage::RECALCULATE_OVERFLOW))
             .unwrap_or(false)
         {
-            return;
+            return self.nodes[node_id].scrollable_overflow;
         }
 
-        self.nodes[node_id].set_transform(self.viewport.scale());
+        let scale = self.viewport.scale_f64();
+
+        let transform = self.nodes[node_id].set_transform(scale as f32);
+
+        let w = self.nodes[node_id].final_layout.size.width as f64 * scale;
+        let h = self.nodes[node_id].final_layout.size.height as f64 * scale;
+        let mut overflow = Rect::new(0.0, 0.0, w, h);
 
         let layout_children = std::mem::take(self.nodes[node_id].layout_children.get_mut());
-        if let Some(lc) = layout_children.as_ref() {
-            for child_id in lc {
-                self.resolve_transforms(*child_id);
+
+        if let Some(ref children) = layout_children {
+            for &child_id in children {
+                let child_rect_in_self = self.resolve_transforms(child_id);
+                overflow = overflow.union(child_rect_in_self);
             }
         }
 
+        self.nodes[node_id].scrollable_overflow = overflow;
         *self.nodes[node_id].layout_children.get_mut() = layout_children;
+
+        let scaled_x = self.nodes[node_id].final_layout.location.x as f64 * scale;
+        let scaled_y = self.nodes[node_id].final_layout.location.y as f64 * scale;
+
+        let full = if let Some(t) = transform {
+            Affine::translate((scaled_x, scaled_y)) * t
+        } else {
+            Affine::translate((scaled_x, scaled_y))
+        };
+
+        full.transform_rect_bbox(overflow)
     }
 
     pub fn resolve_scroll_animation(&mut self) {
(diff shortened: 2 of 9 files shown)


nicoburns:
Thanks for the PR!

> This also needs a redraw call somewhere since live updating the transform doesn't update it in real-time, but I cannot figure out where

Here you've hit on why we don't support this yet. I think we need to introduce a new "overflow" pass between layout and paint so that:

- We can recompute this without recomputing layout
- It's still only recomputed when needed (and ideally incrementally)

Klemen2:
It seems that this requires quite a bit more work than anticipated.
The hit test is also broken now when translate is applied somehow and I'm not entirely sure why it gets shifted by some kind of scale

nicoburns:
Regarding scale, are you rebased on top of https://github.com/DioxusLabs/blitz/commit/0e359fedaf31008ff0a16eebd187245597cbc728 or not? That commit may either have fixed or broken the transform hit testing (it fixes the visual rendering).

But yes, there's quite a bit involved in making this work properly/fast!

Klemen2:
yea I rebased it on the current main. if I understand this correctly I need something like  const ONLY_OVERFLOW: RestyleDamage for damage, enable that damage flag and calculate the transforms

Klemen2:
so I managed to make it smooth, but this is not the best way of doing things as transforms are getting recalculated for every resolve. would there be any better approach than to create a comparisons for styles that trigger relayout and separate comparison for transforms in compute_layout_damage or were you thinking of something else with this 
> "overflow" pass between layout and paint

packages/blitz-dom/src/resolve.rs
```diff
@@ -121,6 +124,26 @@ impl BaseDocument {
         timer.print_times(&format!("Resolve({}): ", self.id()));
     }
 
+    fn resolve_transforms(&mut self, node_id: usize) {
+        if !self.nodes.contains(node_id) {
+            return;
+        }
```
nicoburns, line 131: If you early-return from the this function for nodes that don't have `RestyleDamage::RECALCULATE_OVERFLOW` in their `stylo_element_data` then I think you'll get incrementality (stylo is already computing this for us when properties like transform change).

packages/blitz-dom/src/resolve.rs
```diff
@@ -121,6 +124,26 @@ impl BaseDocument {
...
+
+        self.nodes[node_id].set_transform();
+
+        let children = self.nodes[node_id].children.clone();
```
nicoburns: I think this ought to recurse down `layout_children` rather than `children`.

nicoburns: Also, please avoid cloning here. You can `std::mem::take` and to take the `Vec` and then put it back afterwards (we use this pattern in a bunch of places already).

nicoburns (changes requested):

nicoburns:
There also some issue with this on real webpages:

- Content at top of page is shifted upwards
- Detection for "out of viewport" when scrolling is happening at the wrong offset

https://github.com/user-attachments/assets/2d29b1d2-9998-45dc-9276-30b92793976c

And hit detection is offset for regular (I think not transformed) elements:

https://github.com/user-attachments/assets/3abcc2bf-0ea7-4329-ac55-cf07109a5086

packages/blitz-dom/src/layout/mod.rs
```diff
@@ -450,7 +450,9 @@ impl RoundTree for BaseDocument {
     }
 
     fn set_final_layout(&mut self, node_id: NodeId, layout: &Layout) {
-        self.node_from_id_mut(node_id).final_layout = *layout;
+        let node = self.node_from_id_mut(node_id);
+        node.final_layout = *layout;
+        node.set_transform();
```
nicoburns: This should be unnecessary as a node with "layout" damage will always also get "recalculate_overflow" damage (so the "resolve_transforms" pass should cover this case).

Klemen2:
I was passing element transform instead of the element node transform to children's draw which fixed the issues on webpages

nicoburns:
@Klemen2 I think the underlying issue with the transforms is that the translation (x/y position) is being passed in both the transform and the location/pos parameters, so the translation is being doubled at each step down the tree! It needs to be one or the other. If that was accounted for, then I think passing the "element transform" (as you originally had it) would be correct.

(some special care required around ensuring that the transform-origin for each node is correct)

nicoburns:
Some test cases:

URL: https://github.com/csskit/csskit/discussions

https://github.com/user-attachments/assets/3bfe4d91-bd7f-40fe-9a25-6fa24e17e66a


URL: https://rustweek.org/

https://github.com/user-attachments/assets/5fd28066-eb9d-451d-acb6-7f51eb102939

Klemen2:
the https://rustweek.org/ is now working, I was just applying the transforms in a wrong order, but the github spinner I'm still trying to debug what is actually going on there, the only transform that I can see there is transform: rotate(360deg); in the animation and nothing else

Klemen2:
ah, I figured it out, main branch uses
frame.border_box while I'm calculating it with self.final_layout.size

nicoburns:
It looks to me like the origin for rotations is wrong in some cases (top-left rather than the center of the object?).

My personal website has a similar issue (left: `main`, right: this PR):


https://github.com/user-attachments/assets/a7a6a03b-5cb7-41d2-92d8-01d13a9c96cd

Klemen2:
I believe that I fixed the rendering now, the only thing to fix now is the hit test

nicoburns:
I don't think you actually need to cache the frame to fix the rendering. I think that just multiplying `final_size.width` (and height) by the `scale` would be equivalent to your last commit. That's all that `frame.border_box.width()` (and height) are if you follow through the code.

I definitely haven't fully gotten my head around how the transform needs to interact with the scale yet. And if/how that differs for rendering vs. hit-testing.

Klemen2:
i came to conclusion that the transforms are actually correct now I was yet again applying the transforms in wrong order, your site and the github spinner works, on the https://rustweek.org/ the transforms are technically correct since you see the background does scale correctly but the text is incorrectly scaled / positioned? caching the frame and using it for transform didn't actually result in correct transforms since applying the scale breaks it completely

Klemen2:
this is a minimal reproduction of what is wrong, removing the `display: inline-flex;` from the button does fix it so there is something wrong with displays and scaling, which I'm not sure where they need to be fixed
```rust

fn main() {
    dioxus_native::launch(app);
}

fn app() -> Element {
    rsx! {
        style { {CSS} }
        div { class: "wrapper",
            div { class: "button-group",
                a {
                    class: "button secondary",
                    href: "#week-schedule",
                    "2026 schedule"
                }
                a {
                    class: "button primary",
                    href: "https://2027.rustweek.org",
                    "2027"
                }
            }
        }

    }
}


const CSS: &str = r#"
    .button {
        align-items: center;
        appearance: none;
        border-radius: 1.5rem 0;
        cursor: pointer;
        display: inline-flex;
        font-size: 1rem;
        font-weight: 700;
        line-height: 1;
        min-height: 3rem;
        padding-inline: 1.5rem;
        text-align: center;
        text-decoration: none;
        user-select: none;
        transition-property: filter, scale;
        transition-duration: 0.15s;
    }

    .button.primary {
        background-color: #e74310;
        color: #fff;
    }

    .button.secondary {
        background-color: #729ecb;
        color: #fff;
    }

    .button:visited {
        color: #fff;
    }

    .button:hover {
        filter: brightness(90%);
        scale: 1.2;
    }

    .wrapper {
        display: flex;
        flex-direction: column;
        align-items: left;
        width: fit-content;
    }

    .button-group {
        display: flex;
        flex-wrap: wrap;
        gap: 0.75rem;
        justify-content: center;
    }

"#;
```

nicoburns:
Hmm... it still seems to be getting `transform-origin` wrong:

```html
<!DOCTYPE html>
<style>
  div { width: 50px; height: 50px }
  body { margin: 20px; background: white }
  .black { position: relative; z-index: 0;display: flex; border: 2px solid black; margin: 20px 0; }
  .blue { background: rgba(0, 0, 255, 0.5); }
  .blue:hover { background: rgba(255, 0, 0, 0.5); }
</style>

<div class="black">
    <div class="blue" style="transform: translateY(-20px);"></div>
</div>

<div class="black">
    <div class="blue" style="transform: scale(0.8); transform-origin: center;"></div>
</div>

<div class="black">
    <div class="blue" style="transform: rotate(20deg); transform-origin: center;"></div>
</div>
```

Left: `main`. Right: this PR

Screenshot 2026-05-27 at 15 26 39


At some point we also need to update the bounding-box / scroll area computations to be transform-aware so that we can hit test on a transform node even when it's transform causes it overflow it's parent.

Klemen2:
image
interesting that I cannot reproduce it, maybe I see like 2px offset (firefox left, this commit right), I guess this could be the scale then that is not noticable on smaller screens?

nicoburns:
> image interesting that I cannot reproduce it, maybe I see like 2px offset (firefox left, this commit right), I guess this could be the scale then that is not noticable on smaller screens?

Hmmmm... That's definitely still a little off, but yeah, less than mine! Do you know what your display scale factor is? Could maybe try bumping the sizes in the source html/css and see if that makes it more noticable? (might also help narrow down which style is causing issues...)

nicoburns:
This fixes the transform-origin:

```diff
diff --git a/packages/blitz-dom/src/node/node.rs b/packages/blitz-dom/src/node/node.rs
index a413b3ad..07d51b5e 100644
--- a/packages/blitz-dom/src/node/node.rs
+++ b/packages/blitz-dom/src/node/node.rs
@@ -194,10 +194,10 @@ impl Node {
         }
     }

-    pub fn set_transform(&mut self) {
+    pub fn set_transform(&mut self, scale: f32) {
         self.transform = self.primary_styles().and_then(|s| {
-            let w = self.final_layout.size.width;
-            let h = self.final_layout.size.height;
+            let w = self.final_layout.size.width * scale;
+            let h = self.final_layout.size.height * scale;
             let reference_box = Rect::new(
                 Point2D::new(CSSPixelLength::new(0.0), CSSPixelLength::new(0.0)),
                 Size2D::new(CSSPixelLength::new(w), CSSPixelLength::new(h)),
diff --git a/packages/blitz-dom/src/resolve.rs b/packages/blitz-dom/src/resolve.rs
index 3d58fbad..92b64215 100644
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -137,7 +137,7 @@ impl BaseDocument {
             return;
         }

-        self.nodes[node_id].set_transform();
+        self.nodes[node_id].set_transform(self.viewport.scale());

         let layout_children = std::mem::take(self.nodes[node_id].layout_children.get_mut());
         if let Some(lc) = layout_children.as_ref() {
```

Klemen2:
If you run --example transforms, with scale doesn't really work it actually gets worse or at least it's more visible on my screen, and if you look at that origin dot is also moving with the scale applied, well the origin dot is broken with or without the scale due to scroll offset for some reason. so I am kind of stuck on this for now, I'll see if I figure something out but its definitely more complicated than it seems to be

https://github.com/user-attachments/assets/c4f580c0-ab2c-4542-bbba-54f501d207f6

nicoburns:
I think the origin dot moving is just your "pan" offsets. If I ignore those, then the origin dot seems to scale properly....


https://github.com/user-attachments/assets/4df0f1bd-c427-40af-95c3-ebf7d3a6cecc

Klemen2:
So this is the webview, the boxes are the same which is actually surprising, but notice that the origin dot and text do stays basically on same offset, since the text is ment to be on 0,0. without the scale that origin works, but the boxes are broken so it's either one or the other. I also tried scaling just the origin of the current node and then adding non scaled origin to children which did fix the boxes and origin, but if you have nested transforms it's not great with very unexpected results and as far as I could find it should not need to be differently scaled for children and parent so there is definitely still something wrong, so I'll keep tinkering around and if something works I'll push and we'll see


https://github.com/user-attachments/assets/9b5bf9fb-130c-4fbd-ad1b-78328c61bf8c

Klemen2:
With further testing this is kind of the outcome right now:

webview
webview
this branch
this branch
main branch
main branch

what is broken:
- transform + z-index child
- transform + display: flex child, 
- the bottom box dissapears with too small window size (not sure about main branch since scale is not working)
- transform scale + button text - getting some kind of offset
- feels laggy when updating transforms

what is broken also on main branch
- ToggleBtn component - both gets toggled at the same time / acts like the same component (you'll need to change transform to be on screen)

that offset on your last video is probably due to scroll offset in the last example @nicoburns 

so there is quite a lot that still need to be fixed

nicoburns:
> feels laggy when updating transforms

I'm not seeing this. Are you running release builds?

> the bottom box dissapears with too small window size

This one is straightforward I think. It's just the viewport culling code (that avoids rendering things outside the viewport) not accounting for transforms. Code is here: https://github.com/DioxusLabs/blitz/blob/main/packages/blitz-paint/src/render.rs#L262

> transform + z-index child

For this we may just need to add transform detection to `Node::is_stacking_context_root` in `blitz-dom/src/node/node.rs` (there's a TODO comment for this!)

> transform + display: flex child,

Is this separate from the "button text" issue?

> transform scale + button text - getting some kind of offset

This one is *weird*. I would note that the text seems to be "inline-flex", which means that this is almost certainly going down the "inline box" path rather than the "text span" path.

> ToggleBtn component - both gets toggled at the same time

:/

Klemen2:
> I'm not seeing this. Are you running release builds?

I think this is due to wireless mouse dpi?, release mode is way smoother than debug yes, but like comparing debug main branch to this one there is quite a bit difference
> Is this separate from the "button text" issue?

button is just a default one, without having display or transform set directly to element, so I think so or they can be related

nicoburns:
> I think this is due to wireless mouse dpi?

Ah, yeah possibly we need to do some throttling. I've seen reports about this on the Winit repo, but I don't have a hidpi mouse to test with myself!

Klemen2:
> This one is straightforward I think. It's just the viewport culling code (that avoids rendering things outside the viewport) not accounting for transforms. Code is here: https://github.com/DioxusLabs/blitz/blob/main/packages/blitz-paint/src/render.rs#L262

this actually seems to be a way more complicated problem if I want to keep this check which we probably want?, I'll need to account for children's transforms somewhere where parent size is calculated or something like that since a child can be transformed completely outside an element

> Is this separate from the "button text" issue?

this appears to be a separate issue that is specific for display style attribute & transform

nicoburns:
> > This one is straightforward I think. It's just the viewport culling code (that avoids rendering things outside the viewport) not accounting for transforms. Code is here: https://github.com/DioxusLabs/blitz/blob/main/packages/blitz-paint/src/render.rs#L262
> 
> this actually seems to be a way more complicated problem if I want to keep this check which we probably want?, I'll need to account for children's transforms somewhere where parent size is calculated or something like that since a child can be transformed completely outside an element

I think we'll want to compute the https://drafts.csswg.org/css-overflow-3/#scrollable-overflow-calculation for each box in the same resolve pass where we compute the transforms. Each node ought to return the Axis-Aligned Bounding Box of it's transformed box using https://docs.rs/kurbo/latest/kurbo/struct.Affine.html#method.transform_rect_bbox. Then it's parent can compute it's scrollable overflow from that (or something like that).

We have been using the `content_size` from Taffy, but that's not transform aware.

Klemen2:
right now this mostly works with hit test also working

there also seems to be issues with inputs not applying any transforms if it's directly applied to it

> this appears to be a separate issue that is specific for display style attribute & transform

this appears to be the issue with scroll offset somewhere, e.g the offset will be different if you scroll down on the rustweek page, probably the same issue as draw_inline_layout had in render

Klemen2:
interesting that in browser some elements gets clipped while scrolling on the bottom while the overflow box should be visible, anyway
would you have any idea where some kind of scroll offset would be applied to layout that is changed with display attribute?

nicoburns:
> would you have any idea where some kind of scroll offset would be applied to layout that is changed with display attribute?

I don't think this actually has much to do with `display`. In the example I've just pushed (run with `just open ./examples/assets/inline-flex-transform.html`) it reproduces with flex, grid and block display.

I think this might be coming from combining the transform matrices in the wrong order somehow. And perhaps also to do with the `parent_style_transform` (that I notice is computed separately - suspicious when a parent is transforming correctly, but a child isn't...)

nicoburns:
I think I've fixed it by eliminating the global `pos`/`location` parameters and encoding everything within the transform when rendering. PR into your PR here: https://github.com/Klemen2/blitz/pull/1

(didn't want to push this straight into your branch as I figure this is a potentially controversial change)

nicoburns:
Hmm... there's still an issue with scroll height.

Klemen2:
the transforms of inputs technically do work, but the width doesn't really apply to them which is interesting

nicoburns (approved):
This is looking pretty good to me now. I suspect there are still bugs (hard to catch everything), but I'm not seeing any obvious regressions and there's a small but significant increase in test passes.

nicoburns:
Thanks for all your work on this!

PR #449, opened 2026-06-10
Avoid redraw call when resized to zero size

Klemen2:
with_viewport already calls request_redraw with a safety check 
this should also fix occasional crash when minimizing the window


nicoburns:
Hmm... I'm not against removing duplicate calls to `request_redraw`, but I thought they were harmless because Winit coalesces them into one?

Klemen2:
0.31 does indeed deduplicate automatically. I've changed the title to better describe the change, that occasional crash could have been a driver issue since it used to crash when using use_window().set_minimized(true) and after updating it this doesn't happen on main without changing anything else

PR #453, opened 2026-06-12
Fix display:contents transparency in box-tree construction

UMCEKO:
Stacked on #452 (contains its commit; will rebase once it lands).

`display: contents` is transparent for box generation, but three places in `collect_layout_children` treated the contents element as opaque. Together they made a common Dioxus/Tailwind structure (a `class="contents"` shell between a scroll container and absolutely-positioned pages) lay out as zero-size or content-size boxes. Each fix is covered by a test in `packages/blitz-html/tests/display_contents.rs`:

1. **Classification didn't recurse into contents children.** A container whose in-flow children were all contents wrappers classified as `all_inline` (the existing `TODO: make "all_inline" detection work in the presence of display:contents`) and became an inline root, embedding hoisted *block* children as zero-width inline boxes. The scan now recurses through contents children; the contents node itself casts no vote — including not clearing `all_out_of_flow`, so a container whose only box-generating descendants are out-of-flow takes the out-of-flow path instead of becoming an inline root.

2. **The `DisplayInside::Contents` arm hoisted the wrong generation.** It called `collect_layout_children` on each child, which pushes the child's *layout children* — so leaf children contributed nothing and vanished from the layout tree. Extracted `push_hoisted_children_and_pseudos`, which pushes the children themselves and recurses only through nested contents nodes; the arm also gains `::before`/`::after` support. The same helper now serves the `all_out_of_flow` early-return, which previously pushed the container's direct children — a contents node holding the out-of-flow elements was itself pushed as a layout box, and an `absolute inset:0` child then sized against it instead of the real containing block (content-sized: a real-app page measured 1745x50517 inside a 1004x1251 scroller, leaving nothing scrollable).

3. **Whitespace-only anonymous-block cleanup corrupted the child list.** It removed the anon block with `layout_children.pop()`, assuming it was the most recent entry. Hoisted contents children can land after it, so `pop()` dropped a real child and left the removed anon block's stale id in the list — a slab `invalid key` panic in `resolve_layout_children`. Removal is now by identity.

WPT `css/css-grid` + `css/css-flexbox`: 1154 tests pass before and after.

### DIFF
--- a/packages/blitz-dom/src/layout/construct.rs
+++ b/packages/blitz-dom/src/layout/construct.rs
@@ -203,25 +203,44 @@ pub(crate) fn collect_layout_children(
             // Take children array from node to avoid borrow checker issues.
             let children = std::mem::take(&mut doc.nodes[container_node_id].children);
 
+            // display:contents is transparent for box generation: hoist the
+            // children THEMSELVES (not their layout children) into the
+            // parent, recursing only through nested contents nodes.
             for child_id in children.iter().copied() {
-                collect_layout_children(doc, child_id, layout_children, anonymous_block_id)
+                let child = &doc.nodes[child_id];
+                if child.data.kind() == NodeKind::Comment || child.is_whitespace_node() {
+                    continue;
+                }
+                let child_display = child.display_style().unwrap_or(Display::inline());
+                if matches!(child_display.inside(), DisplayInside::Contents) {
+                    collect_layout_children(doc, child_id, layout_children, anonymous_block_id);
+                } else {
+                    layout_children.push(child_id);
+                }
             }
 
             // Put children array back
             doc.nodes[container_node_id].children = children;
         }
         DisplayInside::Flow | DisplayInside::FlowRoot | DisplayInside::TableCell => {
-            // TODO: make "all_inline" detection work in the presence of display:contents nodes
             let mut all_block = true;
             let mut all_inline = true;
             let mut all_out_of_flow = true;
             let mut has_contents = false;
-            for child in doc.nodes[container_node_id]
+            // display:contents children are transparent for box generation:
+            // their children participate in this container's formatting
+            // context, so classification must recurse into them. Work-stack
+            // in place of plain iteration (reversed so order is preserved,
+            // though classification is order-independent).
+            let mut classify_stack: Vec<usize> = doc.nodes[container_node_id]
                 .children
                 .iter()
                 .copied()
-                .map(|child_id| &doc.nodes[child_id])
-            {
+                .rev()
+                .collect();
+            while let Some(child_id) = classify_stack.pop() {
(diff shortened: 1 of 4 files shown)


Phirios:

The failing job is `Test (macos)` while `Test (linux)` and `Test (windows)` pass. The CI log fails in `blitz-html --test comment_layout`.

I reproduced the failure pattern locally on macOS with the CI-equivalent command:

```sh
```

It failed in `comment_layout`. The focused test binary passes by itself:

```sh
cargo test -p blitz-html --test comment_layout --target aarch64-apple-darwin -- --nocapture
```

And the full CI-equivalent command passes when tests are serialized:

```sh
```

Phirios:

Root cause:

```rust
```

```rust
```

```text
style::driver::traverse_dom
```

How I determined this:

   ```sh
   ```

   ```sh
   ```

Minimal reproducible test:

```rust
use std::thread;

        html,
        DocumentConfig {
            ..Default::default()
        },
    );

    doc.resolve(0.0);

}

#[test]
            </div>
        </div>
    </body></html>"#;

    let mut handles = Vec::new();

    for _ in 0..8 {
            barrier.wait();
            for _ in 0..100 {
            }
        }));
    }

    for handle in handles {
        handle.join().unwrap();
    }
}
```

Run with:

```sh
```

Short-term CI workaround:

```sh
```

UMCEKO:
Rebased the mac ci fix

nicoburns:
@UMCEKO I've refactored this a bunch to:

- Avoid allocating a Vec for classification (use recursive function instread)
- Avoid `O(n)` lookback when pushing the anonymous blocks by eagerly checking for and pushing them every time a node is pushed (also fixes a node ordering bug in your PR I think).

Your review and/or testing of the new version would be appreciated!

UMCEKO:
> @UMCEKO I've refactored this a bunch to:
> 
> - Avoid allocating a Vec for classification (use recursive function instread)
> - Avoid `O(n)` lookback when pushing the anonymous blocks by eagerly checking for and pushing them every time a node is pushed (also fixes a node ordering bug in your PR I think).
> 
> Your review and/or testing of the new version would be appreciated!

Sure! I am currently moving probably going to be able to look into it tmr tho

UMCEKO:
Code quality looks great, I will test the functionality later today on Kopuz

UMCEKO:
> @UMCEKO I've refactored this a bunch to:
> 
> * Avoid allocating a Vec for classification (use recursive function instread)
> * Avoid `O(n)` lookback when pushing the anonymous blocks by eagerly checking for and pushing them every time a node is pushed (also fixes a node ordering bug in your PR I think).
> 
> Your review and/or testing of the new version would be appreciated!

I tested it, looks correct

UMCEKO:
btw i always do tdd on my pr's so if it passes the test its good to go

PR #454, opened 2026-06-12
UA stylesheet: center button text

UMCEKO:
Every major browser's UA sheet applies `text-align: center` to `<button>`; blitz's `default.css` styled buttons but left their text left-aligned. One line + a centering test (the probe is an inline-block embedded box, since plain inline spans don't get a box assigned).


nicoburns (changes requested):
Can we remove the test and the comment? I don't think either are needed for such a small change!

UMCEKO:
Done — dropped both the test and the comment; it's just the `text-align: center` line now. Thanks!

PR #460, opened 2026-06-12
Paint positioned z-index:auto descendants above in-flow content, in tree order

### LINKED ISSUE #458: Later positioned z-index:auto sibling not painted above earlier absolutely-positioned sibling
Per CSS 2.1 Appendix E both siblings sit in paint level 8 ("positioned descendants with z-index: auto") and the later one paints on top; blitz paints the earlier `position: absolute` layer over the later `position: relative` (z-index auto) sibling.

The hero-banner pattern — `absolute inset-0` image/gradient layer followed by a relative content overlay — renders only the image. Workaround: explicit z-index on the overlay.

Repro: `section(position:relative)` > `[div(position:absolute; inset:0; background), div(position:relative, text)]` → text invisible; adding `z-index: 1` to the second child makes it paint.

UMCEKO:
Fixes #458.

Two paint-order fixes in `flush_styles_to_layout`'s `paint_children` sort:

1. `position_to_order` put relative/sticky at the in-flow level (0) and absolute/fixed above it (2), so an earlier abspos sibling painted above a later relative sibling. Per CSS 2.1 Appendix E step 8, all positioned descendants with `z-index: auto` share one paint level and paint in tree order among themselves, above in-flow content. The hero-banner pattern (`absolute inset-0` image layer + relative content overlay) rendered only the image.

2. Flex/grid items ignored `position` entirely (the sort key was just `order`), and z-index was only honored for non-static children — but z-index applies to static flex/grid items too (css-flexbox-1 §painting, css-grid-1 §z-order). The sort key is now (paint level, order-modified position), and static flex/grid items with z-index are hoisted into the stacking context.

Pixel-level regression tests in `packages/blitz-html/tests/paint_order.rs` (render via `anyrender_vello_cpu`, assert the center pixel — adds those as dev-dependencies to blitz-html).

WPT `css/css-grid` + `css/css-flexbox`: **0 regressions**, new passes: `flexbox-items-as-stacking-contexts-002` (its *reference page* depends on fix 1 — it previously "passed" because both pages rendered identically wrong), `flex-item-z-ordering-001/002`, `grid-z-axis-ordering-001/002/003`, `positioned-grid-items-020/021`, `positioned-grid-items-negative-indices-003`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -894,7 +894,10 @@ dependencies = [
 name = "blitz-html"
 version = "0.3.0-alpha.5"
 dependencies = [
+ "anyrender",
+ "anyrender_vello_cpu",
  "blitz-dom",
+ "blitz-paint",
  "blitz-traits",
  "html5ever",
  "tracing",
--- a/packages/blitz-dom/src/layout/damage.rs
+++ b/packages/blitz-dom/src/layout/damage.rs
@@ -549,7 +549,9 @@ impl BaseDocument {
                 let z_index = style.clone_z_index().integer_or(0);
 
                 // TODO: more complete hoisting detection
-                if position != Position::Static && z_index != 0 {
+                // z-index applies to static flex/grid items too
+                // (css-flexbox-1 §painting, css-grid-1 §z-order).
+                if z_index != 0 && (position != Position::Static || is_flex_or_grid) {
                     stacking_context.children.push(HoistedPaintChild {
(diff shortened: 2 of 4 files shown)


nicoburns (commented):
This implementation looks reasonable to me, but tests are failing...

UMCEKO:
Rebased the mac ci fix

PR #461, opened 2026-06-12
Paint overlay scrollbars for scroll containers (feature-gated, default off)

UMCEKO:
blitz previously never painted scrollbars: an `overflow: auto`/`scroll` container gave no visual indication that it scrolls or where within the content the viewport sits.

**Scope correction from the original description:** this PR is **painting only** — the thumb drag/interaction work described earlier was never pushed here (it exists locally and will come as a stacked follow-up PR). Apologies for the confusion; the review was right that the scrollbars in this PR are purely visual.

## What this does

Paints an overlay thumb (no reserved gutter, so no layout change) per overflowing axis: always for `overflow: scroll`, only when the content overflows for `overflow: auto`, never for `hidden`/`clip` (programmatic-only scrolling). Geometry derives from `Layout::scroll_width`/`scroll_height` and `node.scroll_offset`; the thumb is drawn unscrolled, above the clipped content, inside the opacity/filter layer.

Like other overlay scrollbar UIs (macOS, Firefox on Linux), thumbs only appear while the scroll container is hovered or scrolled away from the origin — beyond matching platform conventions, this keeps thumbs out of static reftest screenshots: an always-visible variant regressed 10 WPT overflow reftests (whose reference pages cannot scroll); the hover/scroll rule regresses none.

## Review changes

- **Feature-gated, default off** (as requested): painting sits behind a `scrollbars` cargo feature on `blitz-paint`, mirrored in the `blitz` and `dioxus-native` umbrella crates like the other paint features. Zero code and zero behavior change unless enabled.
- **Closures → functions**: `draw_thumb` / `wants_bar` are now regular (nested) functions with explicit parameters.

## Tests

Pixel-level paint tests in `packages/blitz-html/tests/scrollbars.rs` (render via `anyrender_vello_cpu`, assert pixels); they enable the feature through the dev-dependency.

## Follow-ups (in order)

1. Thumb hit-testing/dragging with hover/active styling (implemented locally; next PR, stacked on this one).
2. css-scrollbars-1 (`scrollbar-color`/`scrollbar-width`) — implemented locally but blocked on enabling those two properties for the servo engine in stylo (their `engine = "gecko"` gate in the property definitions); stylo PR in progress, blitz consumption lands after a stylo release includes it.
3. Track click-to-page, show/hide fade animations.

### DIFF
--- a/packages/blitz-paint/src/render.rs
+++ b/packages/blitz-paint/src/render.rs
@@ -401,6 +401,11 @@ impl<'dom, 'a> BlitzDomPainter<'dom, 'a> {
                                 cx.draw_children(scene, cx.transform);
                             },
                         );
+
+                        // Overlay scrollbars, drawn unscrolled above the
+                        // clipped content.
+                        cx.transform = unscrolled_transform;
+                        cx.draw_scrollbars(scene, overflow_x, overflow_y);
                     },
                 );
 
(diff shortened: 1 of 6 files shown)


Phirios:

The failing job is `Test (macos)` while `Test (linux)` and `Test (windows)` pass. The CI log fails in `blitz-html --test paint_order` (`earlier_abspos_stays_below_static_when_later_in_tree_order_is_static`). I tested the current PR merge ref locally on macOS with:

```sh
cargo test -p blitz-html --test paint_order --target aarch64-apple-darwin -- --nocapture
```

That focused test passes locally. On a related failing PR, the CI-equivalent command:

```sh
```

reproduces a macOS-only failure, but serializing tests makes it pass:

```sh
```

Phirios:

Root cause:

```rust
```

```rust
```

```text
style::driver::traverse_dom
```

How I determined this:

   ```sh
   ```

   ```sh
   ```

Minimal reproducible test:

```rust
use std::thread;

        html,
        DocumentConfig {
            ..Default::default()
        },
    );

    doc.resolve(0.0);

}

#[test]
            </div>
        </div>
    </body></html>"#;

    let mut handles = Vec::new();

    for _ in 0..8 {
            barrier.wait();
            for _ in 0..100 {
            }
        }));
    }

    for handle in handles {
        handle.join().unwrap();
    }
}
```

Run with:

```sh
```

Short-term CI workaround:

```sh
```

packages/blitz-paint/src/render.rs
```diff
@@ -528,6 +533,84 @@ fn convert_rect(rect: &parley::BoundingBox) -> kurbo::Rect {
...
+        }
+        const THUMB_THICKNESS: f64 = 6.0;
+        const THUMB_MARGIN: f64 = 2.0;
+        const MIN_THUMB_LENGTH: f64 = 16.0;
+        const THUMB_COLOR: Color = Color::from_rgba8(128, 128, 128, 178);
+
+        let layout = &self.node.final_layout;
+        let padding_box = self.frame.padding_box;
+
+        let mut draw_thumb = |scroll_extent: f64, scroll_offset: f64, horizontal: bool| {
```
nicoburns: Can we make these closures regular functions instead? I think it makes the code more readable if the parameters are explicitly passed rather than captured. They can still be nested within this function is that is desirable.

nicoburns (changes requested):
@UMCEKO Sorry for the wait on this one. I was a bit conflicted, at least in part because this doesn't actually do what the PR description says it does (interaction is not implemented - the scrollbars are purely visual!). The scorll

I think that's fine (perhaps even preferable) for an initial implementation to keep thing simple. However, I would like to request one thing: that the scrollbars are made optional (default: off), either with a feature flag or runtime setting. That will allow us to mitigate the risk of shipping a relatively immature feature (and may also be useful long term for people who want close control over the visual presentation).

UMCEKO:
> @UMCEKO Sorry for the wait on this one. I was a bit conflicted, at least in part because this doesn't actually do what the PR description says it does (interaction is not implemented - the scrollbars are purely visual!). The scorll
> 
> I think that's fine (perhaps even preferable) for an initial implementation to keep thing simple. However, I would like to request one thing: that the scrollbars are made optional (default: off), either with a feature flag or runtime setting. That will allow us to mitigate the risk of shipping a relatively immature feature (and may also be useful long term for people who want close control over the visual presentation).

ah yeah i later on implemented on top and was going to open a separate pr but ended up merging them together sorry about that

UMCEKO:
@nicoburns I can actually implement the spec properly end to end too, while making this I was kinda hasty as to make our app look and function correctly with blitz but I have more time now, I can do a more proper job with a stylo pr. I will feature gate it.

PR #462, opened 2026-06-13
Make overlay scrollbar thumbs draggable, with hover/active styling

UMCEKO:
Stacked on #461 (contains its commit; will rebase once it lands).

Makes the overlay scrollbar thumbs from #461 interactive:

- Thumb geometry moves from blitz-paint into a shared helper on `Node` (`scrollbar_thumb` / `wants_scrollbar` / `scrollbar_drag_ratio`) so painting and hit testing cannot drift; blitz-paint consumes it.
- Pointer handling gains a `DragMode::ScrollbarDrag` state: on pointerdown the layout ancestor chain of the hit node is walked for a scroll container whose thumb contains the pointer (the hit test returns the content under the overlay thumb; the helper is shared with hover tracking). While dragging, pointer movement maps thumb px to content px via the track ratio and scrolls through the existing clamped `scroll_by` path. Scrollbar drags are not dispatched to the page, and the existing `DragMode` handling suppresses the synthetic click on release.
- The document tracks which thumb the pointer is over (`hovered_scrollbar`, repainting on change), and the painter renders three thumb states: normal, hovered, dragged — a dragged thumb also stays visible if the pointer leaves the container mid-drag.

Tests in `packages/blitz-html/tests/scrollbar_drag.rs` drive the `EventDriver` with synthetic pointer events: thumb drags scroll proportionally and clamp at the track end; content drags don't scroll; hover and drag visibly restyle the thumb.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -894,7 +894,10 @@ dependencies = [
 name = "blitz-html"
 version = "0.3.0-alpha.4"
 dependencies = [
+ "anyrender",
+ "anyrender_vello_cpu",
  "blitz-dom",
+ "blitz-paint",
  "blitz-traits",
  "html5ever",
  "tracing",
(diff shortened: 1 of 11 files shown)


UMCEKO:
Folded into #461 — painting and dragging are one feature; keeping them as two commits in a single PR instead of a 3-deep stack.

PR #464, opened 2026-06-13
Inline roots: apply aspect-ratio to known dimensions at final layout

UMCEKO:
Stacked on #453 (contains its commits). The included layout tests additionally require DioxusLabs/taffy#965 (block aspect-ratio definite height) to pass — draft until both land.

An inline root with `aspect-ratio` and a stretched (known) width never derived its height from the ratio, so e.g. an aspect-square cover containing only inline content sized itself from that content instead. Paired with taffy#965 (the block-layout half of the same gap) and gated to `RunMode::PerformLayout` for the same reason: measure-pass widths are tentative and must not transfer through the ratio.

Layout tests in `packages/blitz-html/tests/image_layout.rs` cover the `` sizing shapes: explicit, percentage, and aspect-ratio-derived container heights, the flex-shelf card shape, and a full-fidelity Tailwind card (where preflight's `img { display: block }` routes through taffy block layout rather than this inline path — both halves are needed).

### DIFF
--- a/packages/blitz-dom/src/layout/construct.rs
+++ b/packages/blitz-dom/src/layout/construct.rs
@@ -203,25 +203,44 @@ pub(crate) fn collect_layout_children(
             // Take children array from node to avoid borrow checker issues.
             let children = std::mem::take(&mut doc.nodes[container_node_id].children);
 
+            // display:contents is transparent for box generation: hoist the
+            // children THEMSELVES (not their layout children) into the
+            // parent, recursing only through nested contents nodes.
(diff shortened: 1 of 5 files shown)


nicoburns:
Not sure if this PR is still relevant, but I've landed your Taffy change in Blitz `main` (released in Taffy v0.12)

UMCEKO:
> Not sure if this PR is still relevant, but I've landed your Taffy change in Blitz `main` (released in Taffy v0.12)

oh nice i forgot about it i'll look into it

PR #465, opened 2026-06-13
Skip pointer-events:none elements in hit-testing

UMCEKO:
## Problem

`Node::hit()` never consulted the `pointer-events` property, so an element with `pointer-events: none` still absorbed pointer hits. This is most visible when a decorative positioned overlay sits above interactive content: per CSS painting order the positioned overlay hit-tests above in-flow siblings, so it swallowed every click and hover meant for the controls beneath it — even though authors set `pointer-events: none` on it precisely to let events through.

## Fix

`hit()` now reads computed `pointer-events`. A `none` element is no longer returned as the hit target itself — neither as a block box nor as the inline text cluster it owns — but its descendants are still tested, so a child that restores `pointer-events: auto` remains hittable (per css-ui-4 inheritance semantics).

## Tests

`packages/blitz-html/tests/pointer_events.rs`:
- a `pointer-events: none` overlay over a button passes the hit through to the button,
- a `none` element falls through to its parent,
- text inside a `none` subtree is not a target,
- a descendant restoring `auto` is targetable.

### DIFF
--- a/packages/blitz-dom/src/node/node.rs
+++ b/packages/blitz-dom/src/node/node.rs
@@ -909,6 +909,7 @@ impl Node {
     /// TODO: z-index
     /// (If multiple children are positioned at the position then a random one will be recursed into)
     pub fn hit(&self, x: f32, y: f32, scale: f64) -> Option<HitResult> {
+        use style::computed_values::pointer_events::T as PointerEvents;
         use style::computed_values::visibility::T as Visibility;
 
         // Don't hit on visbility:hidden elements
@@ -920,6 +921,12 @@ impl Node {
                 return None;
             }
         }
+
+        // pointer-events:none makes this element transparent to hits, but its
+        // descendants are still tested (one may restore pointer-events:auto).
+        let pointer_events_none = self
+            .primary_styles()
+            .is_some_and(|style| style.clone_pointer_events() == PointerEvents::None);
 
         let mut x = x - self.final_layout.location.x + self.scroll_offset.x as f32;
         let mut y = y - self.final_layout.location.y + self.scroll_offset.y as f32;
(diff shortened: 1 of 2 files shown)


nicoburns (approved):

UMCEKO:
i hate macos so much, on it

nicoburns:
It's actually very weird that this is failing only on one platform :/

UMCEKO:
> It's actually very weird that this is failing only on one platform :/

I am going to have my friend who uses mac added to the fork so he can fix it

Phirios:

The failing job is `Test (macos)` while `Test (linux)` and `Test (windows)` pass. The CI log fails in `blitz-html --test pointer_events` (`element_with_pointer_events_none_is_not_a_target`). I tested the current PR merge ref locally on macOS with:

```sh
cargo test -p blitz-html --test pointer_events --target aarch64-apple-darwin -- --nocapture
```

That focused test passes locally. On a related failing PR, the CI-equivalent command:

```sh
```

reproduces a macOS-only failure, but serializing tests makes it pass:

```sh
```

Phirios:

Root cause:

```rust
```

```rust
```

```text
style::driver::traverse_dom
```

How I determined this:

   ```sh
   ```

   ```sh
   ```

Minimal reproducible test:

```rust
use std::thread;

        html,
        DocumentConfig {
            ..Default::default()
        },
    );

    doc.resolve(0.0);

}

#[test]
            </div>
        </div>
    </body></html>"#;

    let mut handles = Vec::new();

    for _ in 0..8 {
            barrier.wait();
            for _ in 0..100 {
            }
        }));
    }

    for handle in handles {
        handle.join().unwrap();
    }
}
```

Run with:

```sh
```

Short-term CI workaround:

```sh
```

nicoburns:
Ah, this makes a lot of sense! I think we probably ought to make `StyleThreading::Sequential` the default, and people can opt-in to `StyleThreading::Parallel` where they know it will work. Otherwise people are going to keep running into this!

UMCEKO:
Do I do it in this branch or will it get merged and we'll do it in a separate pr

nicoburns:
I've pushed the change (https://github.com/DioxusLabs/blitz/commit/648b78018fbb09a811279c9d203685a241662638). So you should just be able to rebase on top of `main`

nicoburns (approved):

PR #466, opened 2026-06-13
Add window-chrome controls to ShellProvider

UMCEKO:
## Problem

A document that draws its own titlebar — a frameless window with an HTML titlebar (drag region + minimize/maximize/close buttons) — had no way to drive the window. `ShellProvider` exposed cursor, title, IME and clipboard, but nothing to move/minimize/maximize/close the window or toggle decorations, so a custom-titlebar app could render the chrome but not operate it.

## Fix

Add window-chrome controls to `ShellProvider` (default no-ops, so existing impls are unaffected):

- `drag_window()` — begin an interactive user-driven move (from a mousedown on a drag region)
- `set_window_minimized` / `set_window_maximized` / `is_window_maximized`
- `set_window_decorations`
- `request_window_close`

`BlitzShellProvider` forwards them to the winit window. Close routes through a new `BlitzShellEvent::CloseWindow { window_id }`, handled identically to `WindowEvent::CloseRequested` (drop the view, exit when the last window closes).

## Tests

- `packages/blitz-traits/tests/shell_provider.rs` — the new methods default to no-ops on a bare provider.
- `packages/blitz-shell/tests/window_controls.rs` — a mock winit `Window` asserts minimize/maximize/decorations/drag forward to the window, and that `request_window_close` emits a `CloseWindow` event for the correct window id.


packages/blitz-shell/src/window.rs
```diff
@@ -604,6 +604,12 @@ impl<Rend: WindowRenderer> View<Rend> {
                                     self.request_redraw();
                                 }
                                 KeyCode::KeyT => self.doc.inner().print_taffy_tree(),
+                                // TEMP (kopuz debugging): dump computed style of
+                                // the route wrapper + its containing block.
+                                KeyCode::KeyY => {
+                                    self.doc.inner().debug_log_node(47);
+                                    self.doc.inner().debug_log_node(49);
+                                }
```
nicoburns: This doesn't look like it ought to be here...?

nicoburns (changes requested):
The changes in this PR seems reasonable, but the tests seem a bit useless seeing as they are just testing a mock implementation (so can we ditch the tests please?)

packages/blitz-shell/src/window.rs
UMCEKO: Oof, i forgot to clean that up, fixing it

UMCEKO:
should be good to go i guess

nicoburns (approved):

PR #475, opened 2026-06-20
Fix templates

### LINKED ISSUE #470: Component merges dynamic attributes to template attributes
In #446 I encountered that weird issue in the transforms example where al ToggleBtn components would get toggled at the same time while sharing nothing

https://github.com/user-attachments/assets/867573f5-d737-47f9-bd72-f118ae46836f

From further testing I believe that the culprit is position: absolute as this is the result of the minimal example bellow, and I believe that the only reason why in the transform examples the ToggleBtn is not drawn on top of each other is due to having different parents.

- position absolute in ToggleBtn

Image

- without position: absolute in children 
(the hit test also fails for some reason, well it fails without any position absolute anywhere so probably unrelated?)

Image

Minimal example:
```rust

fn main() {
    dioxus_native::launch(app);
}

fn app() -> Element {
    rsx! {
        style { {CSS} }
        div {
            position: "absolute",
            top: 0,
            left: 0,
            width: "100%",
            height: "100%",

            ToggleBtn {
                left: "0px",
                width: "100px",
                background: "red"

            }
            ToggleBtn {
                left: "200px",
                width: "100px",
                background: "blue"
            }


        }
    }
}

#[component]
fn ToggleBtn(#[props(extends=GlobalAttributes)] attributes: Vec<Attribute>) -> Element {
    let mut border = use_signal(|| false);

    rsx! {
        button {
            position: "absolute",
            border: if border() { "1px solid black"},
            onclick: move |_| border.toggle(),
             ..attributes,
            "Toggle"
        }
    }
}

const CSS: &str = r#"
html,
body,
main {
    width: 100vw;
    height: 100vh;
    margin: 0;
    overflow: none;
    background: white;
}
"#;
```

Platform: windows 11 (probably unrelated?)
Version: main branch

Klemen2:
Closes: #470
The style is in arc rwlock so every clone of the template was assigned the same style

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -782,11 +782,11 @@ impl BaseDocument {
 
         match &mut data {
             NodeData::Element(elem) | NodeData::AnonymousBlock(elem) => {
-                elem.style_attribute.as_mut().map(|arc| {
+                if let Some(arc) = elem.style_attribute.as_mut() {
                     let read_guard = self.guard().read();
                     let block = arc.read_with(&read_guard);
                     *arc = ServoArc::new(self.guard().wrap(block.clone()));
-                });
+                }
             }
             _ => {}
         }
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
Good catch!

PR #476, opened 2026-06-20
Implement element_coordinates

Klemen2:
I'm not entirely sure if this is the best way of doing this, but it's better than it was

### DIFF
--- a/packages/blitz-traits/src/events.rs
+++ b/packages/blitz-traits/src/events.rs
@@ -442,9 +442,9 @@ pub struct PointerDetails {
 }
 
 #[derive(Copy, Clone, Debug, Default)]
-pub struct ElementData {
-    pub x: f32,
-    pub y: f32,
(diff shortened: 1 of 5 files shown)


packages/blitz-traits/src/events.rs
```diff
@@ -441,6 +441,12 @@ pub struct PointerDetails {
     pub azimuth: f64,
 }
 
+#[derive(Copy, Clone, Debug, Default)]
+pub struct ElementData {
+    pub x: f32,
+    pub y: f32,
+}
+
```
nicoburns, line 449: Could we rename this `ElementCoords`  (or alternatively, just `Point` and make it generic over the inner type)

nicoburns (changes requested):
This mostly looks good to me, but can we rename `ElementData` as that name is pretty generic!

nicoburns (approved):

PR #479, opened 2026-06-27
Workaround notched rendering of borders in Vello by rendering a stroke

rdeaton:
# Reproducer
```html
<!DOCTYPE html>
<!--
  Minimal reproducer: a small `border-radius:50%` element with a thin solid border
  renders with visible seams/notches in the border ring (most clearly at ~9 and
  ~6 o'clock) and an uneven ring width. The same element with no border (a filled
  circle) is smooth, so the artifact is in the BORDER ring construction.

  Render at devicePixelRatio = 2.0 (the artifact is scale-sensitive — see README).
  Drop this into Blitz's `screenshot`/`himg` example, or any blitz-html + Vello path.
-->
<html>
  <head>
    <style>
      html, body { margin: 0; }
      body { background: #9aa0a6; padding: 24px; display: inline-block; }

      /* The disc under test — a board-game stat badge: 44px, thin 1.5px ring. */
      .ring {
        width: 44px;
        height: 44px;
        border-radius: 50%;
        box-sizing: border-box;
        border: 1.5px solid #231b13;
        background: #d64a86;
      }
      /* Control: identical disc with NO border — renders as a smooth circle. */
      .fill {
        width: 44px;
        height: 44px;
        border-radius: 50%;
        background: #d64a86;
      }
    </style>
  </head>
  <body>
    <div class="ring"></div>
    <div class="fill"></div>
  </body>
</html>
```

### DIFF
--- a/packages/blitz-paint/src/kurbo_css/css_box.rs
+++ b/packages/blitz-paint/src/kurbo_css/css_box.rs
@@ -239,7 +239,10 @@ impl CssBox {
     /// path (see [`Self::is_uniform_corner_border`]); this catches the true ellipses
     /// (unequal axes) that a rounded rect can't represent.
     pub fn is_elliptical_border(&self) -> bool {
-        let (rx, ry) = (self.border_box.width() / 2.0, self.border_box.height() / 2.0);
+        let (rx, ry) = (
+            self.border_box.width() / 2.0,
+            self.border_box.height() / 2.0,
+        );
         let is_half = |c: Vec2| (c.x - rx).abs() < 0.01 && (c.y - ry).abs() < 0.01;
         let radii = &self.border_radii;
         is_half(radii.top_left)
--- a/packages/blitz-paint/src/kurbo_css/css_box.rs
+++ b/packages/blitz-paint/src/kurbo_css/css_box.rs
@@ -230,6 +230,42 @@ impl CssBox {
         path
     }
 
+    /// Whether the border box is a full ellipse (which includes a circle): every
+    /// corner radius equals half the box in that axis, i.e. `border-radius: 50%`.
+    /// Such a border is a single continuous curve with no straight edges; it is best
+    /// drawn as a stroked ellipse rather than a filled two-contour annulus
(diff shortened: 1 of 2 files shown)


packages/blitz-paint/src/render.rs
```diff
@@ -881,6 +881,41 @@ impl ElementCx<'_, '_> {
...
+                if (rx - ry).abs() < f64::EPSILON {
+                    let c = kurbo::Circle::new(bb.center(), rx);
+                    scene.stroke(&stroke, self.transform, color, None, &c);
+                } else {
+                    // Ellipse: `Ellipse::to_path` is open, so close it explicitly —
+                    // an open stroked loop leaves a cap gap at the seam.
+                    let mut path =
+                        kurbo::Ellipse::new(bb.center(), (rx, ry), 0.0).to_path(0.1);
+                    path.close_path();
+                    scene.stroke(&stroke, self.transform, color, None, &path);
```
nicoburns: You can pass `Ellipse` (anything that implements `kurbo::Shape`) directly to `scene.stroke`. No need to convert to path first.

nicoburns (changes requested):
This seems like a good change, but can we generalise it to detecting rounded rectangles with uniform color, uniform border width, and where the x and y radii in each corner are equal? (and then stroke the rounded rect)

I think this will probably be faster than the fallback annulus approach for rects and rounded rects too, and I think it might make the detection logic a bit simpler!

packages/blitz-paint/src/render.rs
rdeaton: iirc, the issue here was that the ellipse's path wasn't closed (like Circle is) which caused a rendering artifact, but I'll double check as I try the suggested other fixes.

rdeaton: If it did, I couldn't reproduce it after switching to the rounded rects, so indeed dropped it.

rdeaton:
@nicoburns ready for round two, thanks.

PR #489, opened 2026-07-03
Implement scrollbars end to end

UMCEKO:
Implements scrollbar support end to end [as per the w3 spec](https://www.w3.org/TR/css-scrollbars-1/) and compared against Blink/Chromium for any edge cases.

Generated via Claude Fable, code quality reviewed manually. PR Desc was handwritten.

## Changes made:

- Scroll containers now paint an overlay scrollbar thumb (hidden at rest, shown on hover/scroll)
- Thumbs can be dragged to scroll, with hover/drag highlighting
- `scrollbar-color` and `scrollbar-width` (`thin`/`none`) are supported
- Everything is behind a `scrollbars` cargo feature, **off by default**

## Notes

- Awaiting servo/stylo#413 to get merged to revert Cargo.toml and Cargo.lock, stylo source was switched so tests pass and I could test.

## Out of scope / deliberately skipped

- Fade in/out animations and hover thickening needs repaint timer so they were skipped to be implemented in a separate pr
- Scrollbars on the viewport itself were skipped because blitz does not paint those at all yet.
- RTL and vertical writing modes - blitz doesn't support RTL scroll ranges yet.
- Thumb hit-testing inside CSS transforms, pointer-math limitations were skipped, added a fixme in code so we don't forget about it.
- Classic space reserving scrollbars, `scrollbar-gutter`

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -249,6 +249,9 @@ pub struct BaseDocument {
     pub(crate) drag_mode: DragMode,
     /// The scrollbar thumb currently under the pointer, if any
     pub(crate) hovered_scrollbar: Option<crate::node::ScrollbarRef>,
+    /// When each scroll container's overlay scrollbars were last shown
+    /// (scrolled, or the pointer left the thumb); drives their fade-out
+    pub(crate) scrollbar_activity: HashMap<usize, Instant>,
     /// Whether and what kind of scroll animation is currently in progress
     pub(crate) scroll_animation: ScrollAnimationState,
 
@@ -463,6 +466,7 @@ impl BaseDocument {
             click_count: 0,
             drag_mode: DragMode::None,
             hovered_scrollbar: None,
+            scrollbar_activity: HashMap::new(),
             scroll_animation: ScrollAnimationState::None,
             text_selection: TextSelection::default(),
         };
@@ -1339,6 +1343,42 @@ impl BaseDocument {
         }
     }
 
+    /// The current opacity of `node_id`'s overlay scrollbars. They show at
+    /// full opacity on scroll and fade out after a delay (Chromium's overlay
+    /// timings); the pointer resting on a thumb, or dragging it, holds them
+    /// visible.
+    pub fn scrollbar_opacity(&self, node_id: usize) -> f32 {
+        let interacting = |scrollbar: &crate::node::ScrollbarRef| scrollbar.node_id == node_id;
+        if self.hovered_scrollbar.as_ref().is_some_and(interacting)
+            || self
+                .scrollbar_drag_target()
+                .as_ref()
+                .is_some_and(interacting)
+        {
+            return 1.0;
+        }
+        self.scrollbar_activity.get(&node_id).map_or(0.0, |last| {
+            crate::node::scrollbar::opacity_at(last.elapsed())
+        })
+    }
+
+    /// Show `node_id`'s overlay scrollbars at full opacity and restart their
+    /// fade-out delay.
+    pub(crate) fn show_scrollbars(&mut self, node_id: usize) {
+        if cfg!(feature = "scrollbars") {
+            self.scrollbar_activity.insert(node_id, Instant::now());
+        }
+    }
+
+    /// Whether any overlay scrollbars are awaiting or animating their
+    /// fade-out (so frames must keep rendering until they finish).
+    fn scrollbars_animating(&self) -> bool {
+        use crate::node::scrollbar::{FADE_DELAY, FADE_DURATION};
+        self.scrollbar_activity
+            .values()
+            .any(|last| last.elapsed() < FADE_DELAY + FADE_DURATION)
+    }
+
     /// [`hit`](Self::hit), also resolving the innermost overlay scrollbar
     /// thumb under the point (shares the traversal, so it costs nothing
     /// extra).
@@ -1364,10 +1404,24 @@ impl BaseDocument {
 
     pub fn set_hover_to(&mut self, x: f32, y: f32) -> bool {
         let (hit, hovered_scrollbar) = self.hit_with_scrollbar(x, y);
+        // A faded-out thumb is not interactive: pointer moves never fade
+        // overlay scrollbars back in (only scrolling shows them).
+        let hovered_scrollbar =
+            hovered_scrollbar.filter(|scrollbar| self.scrollbar_opacity(scrollbar.node_id) > 0.0);
         // Scrollbar-thumb hover is part of hover state: track it here so a
         // pointer crossing a thumb restyles it even when the hit node (the
         // content under the overlay thumb) is unchanged.
         let scrollbar_changed = hovered_scrollbar != self.hovered_scrollbar;
+        if scrollbar_changed {
+            // Entering a thumb restores full opacity mid-fade; leaving one
+            // restarts the fade-out delay.
+            for scrollbar in [self.hovered_scrollbar, hovered_scrollbar]
+                .into_iter()
+                .flatten()
+            {
+                self.show_scrollbars(scrollbar.node_id);
+            }
+        }
         self.hovered_scrollbar = hovered_scrollbar;
         let hover_node_id = hit.map(|hit| hit.node_id);
         let new_is_text = hit.map(|hit| hit.is_text).unwrap_or(false);
@@ -1530,6 +1584,7 @@ impl BaseDocument {
             | self.subdoc_is_animating
             | has_custom_widgets
             | (self.scroll_animation != ScrollAnimationState::None)
+            | self.scrollbars_animating()
     }
 
     /// Update the device and reset the stylist to process the new size
@@ -1777,8 +1832,13 @@ impl BaseDocument {
             dispatch_event(DomEvent::new(node_id, DomEventData::Scroll(event)));
         }
 
+        let parent = node.parent;
+        if has_changed {
+            self.show_scrollbars(node_id);
+        }
+
         if bubble_x != 0.0 || bubble_y != 0.0 {
-            if let Some(parent) = node.parent {
+            if let Some(parent) = parent {
                 return self.scroll_node_by_has_changed(parent, bubble_x, bubble_y, dispatch_event)
                     | has_changed;
             } else {
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -247,9 +247,8 @@ pub struct BaseDocument {
     pub(crate) click_count: u16,
     /// Whether we're currently in a text selection drag (moved 2px+ from mousedown)
     pub(crate) drag_mode: DragMode,
-    /// The scrollbar thumb (scroll container id, is-horizontal) currently
-    /// under the pointer, if any
-    pub(crate) hovered_scrollbar: Option<(usize, bool)>,
+    /// The scrollbar thumb currently under the pointer, if any
+    pub(crate) hovered_scrollbar: Option<crate::node::ScrollbarRef>,
     /// Whether and what kind of scroll animation is currently in progress
     pub(crate) scroll_animation: ScrollAnimationState,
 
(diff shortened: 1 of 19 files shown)


UMCEKO (commented):
Fixed some parts where some functionality was duplicated, rest of the code quality lgtm but idk the conventions of this codebase so there might be some things I could've missed.

packages/blitz-paint/src/render.rs
```diff
@@ -563,6 +571,191 @@ fn convert_rect(rect: &parley::BoundingBox) -> kurbo::Rect {
...
+            fn lin(c: f32) -> f32 {
+                if c <= 0.04045 {
+                    c / 12.92
+                } else {
+                    ((c + 0.055) / 1.055).powf(2.4)
+                }
+            }
+            let [r, g, b, _] = color.components;
+            0.2126 * lin(r) + 0.7152 * lin(g) + 0.0722 * lin(b)
+        }
```
nicoburns: This is implemented in the color crate as `OpaqueColor::relative_luminance`. We should use that.

packages/blitz-paint/src/render.rs
```diff
@@ -563,6 +571,191 @@ fn convert_rect(rect: &parley::BoundingBox) -> kurbo::Rect {
...
+
+        /// WCAG contrast ratio between two relative luminances.
+        fn contrast_ratio(l1: f32, l2: f32) -> f32 {
+            let (hi, lo) = if l1 > l2 { (l1, l2) } else { (l2, l1) };
+            (hi + 0.05) / (lo + 0.05)
+        }
```
nicoburns: This or a very similar function exists in `border.rs`. Could we maybe move all the color helpers into `color.rs` and share them across `blitz-paint`?

nicoburns:
Are the scrollbars supposed to be so thin? Possibly scale factor ought to be accounted for? (or possibly you've just chose a very thin width):

Screenshot 2026-07-03 at 20 51 51

packages/blitz-paint/src/render.rs
UMCEKO: Sure

packages/blitz-paint/src/render.rs
UMCEKO: Oop, i missed that, doing that

UMCEKO:
> Are the scrollbars supposed to be so thin? Possibly scale factor ought to be accounted for? (or possibly you've just chose a very thin width):
> 
> Screenshot 2026-07-03 at 20 51 51

let me compare it against chromium brb

UMCEKO:
Yeah chromium's thumbnail was 10 px thick, min length was 32 px, fixed accordingly in new commit to be in par with chromium.

nicoburns:
A suggestion for being able to land this sooner:

- Define our own versions of the `ScrollbarWidth` and `ScrollbarColor` types from Stylo (the one's you're enabling in your PR) in `blitz-dom`.
- Use them in this PR in place of the Stylo types.
- Where you are reading the style from Stylo, comment out that code and hardcode a default value.

Then we could land most of the implementation, and just keep a small diff in a PR that switches in the real style types.

(I expect we'll be able to land the PR in Stylo relatively soon, but Stylo releases are monthly and one just went out today so it'll probably be ~a month until that filter down into Blitz)

UMCEKO:
> A suggestion for being able to land this sooner:
> 
> * Define our own versions of the `ScrollbarWidth` and `ScrollbarColor` types from Stylo (the one's you're enabling in your PR) in `blitz-dom`.
> * Use them in this PR in place of the Stylo types.
> * Where you are reading the style from Stylo, comment out that code and hardcode a default value.
> 
> Then we could land most of the implementation, and just keep a small diff in a PR that switches in the real style types.
> 
> (I expect we'll be able to land the PR in Stylo relatively soon, but Stylo releases are monthly and one just went out today so it'll probably be ~a month until that filter down into Blitz)

I'll look into it tmr

UMCEKO:
Nvm couldn't sleep, blitz-dom now has its own ScrollbarWidth/ScrollbarColor mirrors. Stylo reads are left as todos.

packages/blitz-dom/src/events/pointer.rs
```diff
@@ -46,6 +46,15 @@ pub(crate) struct PanSample {
     pub(crate) dy: f32,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub(crate) struct ScrollbarDragState {
+    /// The scroll container whose thumb is being dragged
+    pub(crate) node_id: usize,
+    pub(crate) horizontal: bool,
```
nicoburns: Optional: Perhaps this should be an enum? `taffy::AbsoluteAxis` would probably work.

packages/blitz-dom/src/node/node.rs
```diff
@@ -494,6 +494,30 @@ pub enum NodeKind {
     Comment,
```
nicoburns, line 494: Could we move all the new scrollbar methods into a dedicated `scrollbar.rs` in the `node` directory? (`node.rs` is getting kinda big...)

packages/blitz-dom/src/node/node.rs
```diff
@@ -1110,6 +1134,151 @@ impl Node {
...
+        match overflow {
+            Overflow::Scroll => true,
+            Overflow::Auto => scroll_extent > 0.5,
+            _ => false,
+        }
+    }
+
+    /// The scrollport (padding box) in (unscaled) CSS px relative to the
+    /// node's border-box origin. Taffy has content-box helpers but none for
+    /// the padding box.
```
nicoburns: We probably ought to add such helpers to Taffy at some point (but this is low priority).

(I'm also hoping to unify Taffy and Kurbo's geometry type representations at some point (https://github.com/linebender/kurbo/pull/513))

packages/blitz-dom/src/document.rs
```diff
@@ -248,6 +247,9 @@ pub struct BaseDocument {
     pub(crate) click_count: u16,
     /// Whether we're currently in a text selection drag (moved 2px+ from mousedown)
     pub(crate) drag_mode: DragMode,
+    /// The scrollbar thumb (scroll container id, is-horizontal) currently
+    /// under the pointer, if any
+    pub(crate) hovered_scrollbar: Option<(usize, bool)>,
```
nicoburns: Can me make this a struct please? It can also live in `scrollbars.rs`.

nicoburns: Or, could this be de-duplicated with the `hover_node_id` and we just store `Option<AbsoluteAxis>` here?

(or do you think that will be confusing?)

packages/blitz-paint/src/render.rs
```diff
@@ -563,6 +571,130 @@ fn convert_rect(rect: &parley::BoundingBox) -> kurbo::Rect {
...
+        use blitz_dom::node::ScrollbarColor;
+        let (custom_thumb, custom_track) = match self.node.scrollbar_color() {
+            ScrollbarColor::Auto => (None, None),
+            ScrollbarColor::Colors { thumb, track } => {
+                (Some(thumb.as_srgb_color()), Some(track.as_srgb_color()))
+            }
+        };
+
+        let drag_target = self.context.dom.scrollbar_drag_target();
+        let hovered_thumb = self.context.dom.hovered_scrollbar();
```
nicoburns: Like the root element, I think these could probably be resolved once rather than for every element.

packages/blitz-paint/src/render.rs
```diff
@@ -563,6 +571,130 @@ fn convert_rect(rect: &parley::BoundingBox) -> kurbo::Rect {
...
+
+        // scrollbar-color doesn't affect overlay visibility: persistence is
+        // UA policy, not author styling.
+        let node_id = self.node.id;
+        let is_drag_target = matches!(drag_target, Some((target_id, _)) if target_id == node_id);
+        if !self.node.is_hovered()
+            && !is_drag_target
+            && self.node.scroll_offset.x == 0.0
+            && self.node.scroll_offset.y == 0.0
```
nicoburns: I'm not 100% sure about this "scrollbar are visible if the node has been scrolled" logic (seems non-standard?). But I'm also ok with it, and not sure what would be better.

packages/blitz-dom/src/events/pointer.rs
```diff
@@ -138,6 +149,48 @@ impl PanState {
...
+            };
+            // absolute_position subtracts the node's own scroll offset
+            // (it maps content coordinates); compensate to get the
+            // border-box origin.
+            //
+            // FIXME: absolute_position ignores CSS transforms, so the hit
+            // region is wrong inside transformed ancestors (as is blitz's
+            // other pointer math). Wants a shared transform-aware
+            // page->local conversion, not a scrollbar-local fix.
+            let origin = node.absolute_position(0.0, 0.0);
```
nicoburns: Seems a bit unfortunate that we're re-resolving the position up the entire tree for every node (`O(n^2)` complexity). But I think I'm happy to fix this later.

I'm thinking that the best way to optimise this may be:

- Store a `hover_node_path: Vec<usize>` on the BaseDocument, rather than just `hover_node_id: usize`
- Then we could recurse down the tree (applying transforms as we go) rather than up it.
- (we could even store `x`/`y` coords in local node space for each ancestor in the tree?)

Alternatively, perhaps resolving the scrollbar could be integrated with `hit`  method?

Not sure what the best way to balance complexity and efficiency is here. But given this could end up being called very frequently (every mouse move) we need to be conscious of performance

nicoburns (changes requested):
This mostly looks good, but I have some comments.

packages/blitz-dom/src/events/pointer.rs
UMCEKO: I'll resolve this by looking up the chromium implementation and replicating it

packages/blitz-dom/src/events/pointer.rs
UMCEKO: Done.

packages/blitz-dom/src/node/node.rs
UMCEKO: Done.

packages/blitz-dom/src/document.rs
UMCEKO: I don't think they can be separated, they are different nodes. The hit test returns the content under the overlay thumb whilst the thumb belongs to an ancestor scroll container so hovered_scrollbar isn't really derivable from hover_node_id. This approach mirrors blink (https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/layout/hit_test_result.h), which keeps them independent members and EventHandler tracks last_scrollbar_under_mouse_ separately from the hovered-node chain (https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/input/event_handler.h).

UMCEKO: They do share the traversal now tho, set_hover_to resolves both in one descent, and the O(depth^2) upward walk is eliminated.

packages/blitz-paint/src/render.rs
UMCEKO: Done.

packages/blitz-dom/src/node/node.rs
UMCEKO: Yeah taffy doesn't have any helpers for the padding box. Happy to upstream a helper to taffy if you'd like tho.

packages/blitz-paint/src/render.rs
UMCEKO: Resolved this properly by replicating Chromium's overlay visibility.

- scrolling shows the bars
- they hold 500ms then fade over 200ms
- hovering/dragging the thumb holds them
- faded out thumbs don't capture pointer events or fade back in on pointer moves

Opacity is a pure function of time since last scroll activity, so there's no timer machinery, is_animating() keeps frames coming while a fade is pending, same as the fling animation.

packages/blitz-dom/src/events/pointer.rs
UMCEKO: Your proposed "resolving the scrollbar could be integrated with hit method" is the way chromium did it, so I went with it. Resolved.

packages/blitz-dom/src/document.rs
UMCEKO: > Can me make this a struct please? It can also live in `scrollbars.rs`.

Done.


> Or, could this be de-duplicated with the `hover_node_id` and we just store `Option<AbsoluteAxis>` here?
> 
> (or do you think that will be confusing?)

kept it separate from hover_node_id, but they are no longer resolved separately, one hit-test descent now produces both.

packages/blitz-paint/src/render.rs
UMCEKO: Done

nicoburns (approved):

PR #491, opened 2026-07-05
JavaScript support

nicoburns:
This is an AI-generated implementation of JavaScript support for Blitz using the Boa JS engine. I have not yet reviewed the code.

Generated with Claude Fable XHigh

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/32034316613).</sub>


devarda:
Hello there, I have a scope question. Do you envision networking Web APIs like `fetch()`, `WebSocket`, and network-backed document/script loading via blitz-net living inside blitz-script eventually, or is that embedder territory? Thanks

P.S. I've been building an embedder on this branch for the past while. A React 18 + TS apps and a small real-world web app run end-to-end on blitz-script.  (with a handful of local DOM additions: classList, closest, dataset, etc. — happy to send those as PRs if useful).

nicoburns:
I would definitely be interested to see a variant of this PR that can run React apps!

That being said, I think this PR has architectural issues, and probably isn't mergeable in this form (+I still haven't gone through the whole thing in detail yet).

> Do you envision networking Web APIs like fetch(), WebSocket, and network-backed document/script loading via blitz-net living inside blitz-script eventually, or is that embedder territory?

I'd guess we'd at least want to supply an implementation of these, but keep them optional. Whether it's in the same crate or a separate crate I don't know.

devarda:
Yeah, I hear you! Thanks!

Here's the PR link with my additions: https://github.com/nicoburns/blitz/compare/js-engine...devarda:blitz:react-dom-apis

Here's my demo project: https://github.com/devarda/react-blitz-demo
Here's my fork of this that has the PR above: https://github.com/devarda/blitz/tree/react-dom-apis

Mind you that my code was AI generated as well. I've been checking the outcomes and not the code. I haven't written any code for over 2 years now... Idk what to think of it xD 

Anyway, perhaps you can have your AI take a look at my AI's commits. 

I look forward to the end results of your work here! 

Best.

Screenshot 2026-07-09 at 7 41 23 PM

nicoburns:
This is very cool! It's great news that React will run with a relatively small set of changes (I was expecting more to be required tbh). I imagine there are parts of React that still don't work here. But if it's usable then that's a really good start! Probably I should rename the example, so you don't have to run an example called `preact_script` when you're running React code!

> Anyway, perhaps you can have your AI take a look at my AI's commits.

Alas, I am still committed to having human review on everything that goes into Blitz (which is also partly why this PR is still a draft - I haven't had a chance to review it yet either). Luckily the diff from your changes actually seems to be relatively small/manageable (~1k LoC), so I think I can review. I had a quick skim, and it looks pretty reasonable.

PR #492, opened 2026-07-06
Drag & drop

Klemen2:
Stacked on top of: #614

This PR adds the following
### Drag events:
- [x] `dragenter` event
- [x] `dragover` event
- [x] `dragleave` event
- [x] `drop` event
- [x] `dragstart` event
- [x] `drag` event
- [x] `dragend` event
### External drags:
- [x] dragging files
- [x] dragging text (plain/html/rtf)
- [x] ability to downcast data transfer to handle platform specific data types
### Internal drags
- [x] dragging out of the window
- [x] dragging text (plain)
### Shell
- [x] `set_datatransfer`

Not included in this pr:  dragging text (html), dragging images, dragging links, dragging input/textbox text, dropping to input/textbox, cursor icon, drop effect based on target

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/34267638756).</sub>


packages/blitz-traits/src/events.rs
```diff
@@ -441,11 +497,166 @@ impl DomEventData {
...
+        matches!(self, Self::File)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct BlitzDragDataItem {
+    pub kind: BlitzDragDataItemKind,
+    pub format: String,
+    pub data: String,
+}
```
Klemen2: I feel like this should be a trait instead, or at least the data?

nicoburns:
Will take a proper look but:

> This is blocked on most likely next winit release as it's using the new dnd api

Just wanted to say that this seems like a very sensible choice. I'm anticipating that we'll probably get a new Winit beta once the DnD PR lands.

Klemen2:
No need to rush, I've only tested the incoming dnd events and not everything is implemented yet, but it's a great start. 
I also think that I need to change how DataTransferReceived is handled since i believe it can come after the drop is handeled

packages/dioxus-native/src/config.rs
```diff
@@ -32,6 +33,7 @@ impl Default for Config {
         Self {
             window_attributes,
             font_ctx: None,
+            enable_dnd: false,
```
nicoburns: Why is this configurable? (when would you ever want it disabled?- you can just not handle the drags?)

packages/blitz-traits/src/events.rs
```diff
@@ -441,11 +496,166 @@ impl DomEventData {
...
+
+#[derive(Clone, Debug, Default)]
+pub struct BlitzDataTransfer {
+    pub items: Vec<BlitzDragDataItem>,
+    pub effect_allowed: Option<BlitzDragOperations>,
+    pub drop_effect: BlitzDragOperation,
+    pub mode: BlitzDragDataStoreMode,
+}
+
+pub type BlitzDragId = i64;
```
nicoburns: `i64` seems weird. Unless there's a reason to use a signed int can we make this `u64` please?

nicoburns (commented):
I took a look because I was curious and have a left a couple of notes.

And a general comment that:

> Drag and Drop actually has three distinct use cases: [dragging elements within a page](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Kanban_board), dragging data out of a page, and [dragging data into a page](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop). They have subtly different requirements and implementations.

(https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API)

And this only currently only implements "dragging data into a page". We don't necessarily need an implementation of all 3 at once, but it's at least worth thinking about when designing the API (and I'd personally be kinda tempted to try to handle "dragging elements within a page" too, as I think that might end up changing the design somewhat).

packages/dioxus-native/src/config.rs
Klemen2: winit 0.30 had an option to disable that, or at least on windows, I can remove it. it probably should have a setting about what kind of files it should handle / what kind of dnd action it should handle as you may not want the app to handle everything if not needed instead

packages/blitz-traits/src/events.rs
Klemen2: winit uses i64 as inner DataTransferId so I just used that here, so I can change this as well

Klemen2:
I think that the enter/over/leave/drop are now handled correctly (needs #496 to be fixed for bubbling) 

I think that the structure can support "dragging elements within a page" but I haven't come to that part yet, also dragable=true needs to handle the event data automagically so yea fun stuff

Klemen2:
So I tried implementing dragstart/drag/dragend, but for whatever reason ondragstart is not even getting triggered... am I missing something somewhere? it gets to  self.handler.handle_event in the driver, but not to the actual event, I don't think the implementation is correct yet, especially drag/dragend so thats still wip

nicoburns:
@Klemen2 You need to detect and send `dragstart` (and set blitz-dom into a "dragging" state) yourself! See `pointer.rs` where we already detect drags for text selection and pan-to-scroll on mobile.

EDIT: well that's for dragging DOM elements with `draggable=true`. For external drags, you should probably be firing `DragStart` when first encountering a new `DragEnter`. Winit DragEvent(enter window)  != web dragenter (enter droppable node).

Klemen2:
I did some changes and now you can d&d text within the app, haven't implemented the outgoing drags yet, but ondragstart is still not triggering on draggable (implementation is inside `pointer.rs`)

> For external drags, you should probably be firing DragStart

as far as I can tell, dragstart only fires on elements with draggable=true / draggable types (text, images, link)  from where the drag started, and doesn't fire at all for external drags

Klemen2:
I think that this is mostly ready, (todo: cursor icon, dragging images, inputs, links)
I don't really know how should i be implementing get_datatransfer for the shell, std::mpsc? or futures oneshot? 
Should internal drags be registered with OS (using event loop) when drag starts or when cursor leaves the window (the way its now)? - I think I answered this myself with just asking this question

nicoburns:
Testing this on macOS I'm getting two files when dropping a single file onto the drop zone

Klemen2:
> Testing this on macOS I'm getting two files when dropping a single file onto the drop zone

Interesting, I'm not experiencing this on windows

packages/blitz-traits/src/events/datatransfer.rs
```diff
@@ -0,0 +1,306 @@
...
+            return Err("setData is only allowed during dragstart".into());
+        }
+        let format = normalize_format(format);
+        self.items.set_data(&format, data)
+    }
+
+    pub fn clear_data(&mut self, format: Option<&str>) -> Result<(), String> {
+        if self.mode.can_write() {
+            return Err("clearData is only allowed during dragstart".into());
+        }
```
nicoburns: This check is inverted

packages/blitz-dom/src/events/driver.rs
```diff
@@ -144,12 +144,44 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
...
+                DomEventData::DragLeave(event.clone()),
+            ));
+        }
+
+        if let Some(target) = hover_node_id {
+            self.handle_dom_event(DomEvent::new(
+                target,
+                DomEventData::DragEnter(event.clone()),
+            ));
+        }
```
nicoburns, line 174: DragEnter should fire before DragLeave. See https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations#dragging_over_elements_and_specifying_drop_targets

packages/blitz-dom/src/events/driver.rs
```diff
@@ -254,6 +301,38 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
                 self.run_default_action(&mut dom_event);
                 self.process_queue();
             }
+            UiEvent::DragEnter(data) => {
+                self.handle_dom_event(DomEvent::new(target, DomEventData::DragEnter(data)))
+            }
```
nicoburns, line 311: `UIEvent::DragEnter` and `UIEvent::DragLeave` need some attention. They don't map cleanly to the dom events of the same name. In particular, when first entering a window, (dom) dragenter currently fires twice.

packages/dioxus-native-dom/src/events.rs
```diff
@@ -603,6 +610,175 @@ pub fn synthetic_click_event(node: &Node, modifiers: Modifiers) -> Box<dyn Any>
...
+    fn byte_stream(
+        &self,
+    ) -> Pin<
+        Box<
+            dyn futures_util::Stream<
+                    Item = Result<blitz_traits::net::Bytes, dioxus_core::CapturedError>,
+                >
+                + 'static
+                + Send,
+        >,
```
nicoburns: Can we maybe import the types with `use` so this formats better?

packages/blitz-shell/src/window.rs
```diff
@@ -823,10 +916,142 @@ impl<Rend: WindowRenderer> View<Rend> {
...
+                                succesful_fetch = true;
+                                self.pending_data_transfers.push(serial);
+                            },
+                            Err(err) => {
+                                #[cfg(feature = "tracing")]
+                                tracing::error!("{}", err);
+                            },
+                        }
+                    ControlFlow::Continue(())
+                } );
```
nicoburns: Data should probably only be fetched on Drop

nicoburns (changes requested):

packages/blitz-shell/src/window.rs
Klemen2: this requires changing https://github.com/DioxusLabs/dioxus/blob/main/packages/html/src/data_transfer.rs `fn files` and `fn get_data` to `Pin<Box<dyn Future` as this needs to wait for the future event loop to be executed.

packages/blitz-dom/src/events/driver.rs
Klemen2: When trying to debug this I found that the dragenter triggers on node 72 in this instance, which first appears in nodes_4 and is then an orphaned node afterwards so I don't think that this is necessarily the problem just for drag events but something to do with how anonymous blocks are created/handled. It is also worth mentioning that at the start there are 65 (24 anonymous) nodes and at the end there are 93 (52 anonymous) while just performing a drag and nothing else.
image

[nodes_1.json](https://github.com/user-attachments/files/30340918/nodes_1.json)
[nodes_2.json](https://github.com/user-attachments/files/30340919/nodes_2.json)
[nodes_3.json](https://github.com/user-attachments/files/30340922/nodes_3.json)
[nodes_4.json](https://github.com/user-attachments/files/30340923/nodes_4.json)
[nodes_5.json](https://github.com/user-attachments/files/30340924/nodes_5.json)
[nodes_6.json](https://github.com/user-attachments/files/30340926/nodes_6.json)
[nodes_7.json](https://github.com/user-attachments/files/30340928/nodes_7.json)
[nodes_8.json](https://github.com/user-attachments/files/30340931/nodes_8.json)
[nodes_9.json](https://github.com/user-attachments/files/30340932/nodes_9.json)

packages/blitz-shell/src/window.rs
Klemen2: nevermind I made it so that the drop event gets sent after all data is fetched

Klemen2:
~~@nicoburns how do I fix this returning 33 for drag start since the drag start event is 34?~~
https://github.com/DioxusLabs/blitz/blob/f6653829d8ab8f61542aeacbb0ee17946808706c/packages/dioxus-native-dom/src/dioxus_document.rs#L268

~~I don't think that doing this is the solution~~
```rust
let mut event_kind_idx = event.data.discriminant() as usize;
match event.data {
    DomEventData::DragStart(_) => event_kind_idx += 1,
    _ => {}
}
```
~~the custom draggable do work after doing this~~

Fixed it

packages/blitz-dom/src/events/driver.rs
Klemen2: is there a reason why this gets called pretty much all the time https://github.com/DioxusLabs/blitz/blob/f6653829d8ab8f61542aeacbb0ee17946808706c/packages/blitz-dom/src/layout/construct.rs#L318  just by hovering over which calls `collect_complex_layout_children` which calls `push_wrapped` which calls `create_anonymous_block` all the time that create new anonymous nodes without ever getting reused ? 
which then get in a way of the events since the event is bubbling and it works as intended, since it entered a new node

nicoburns: If there's a pseudo-element created on `:hover` by the `content` property then we need to re-collect children. We should try to re-use anonymous elements. But we also shouldn't rely on their IDs being stable. They should be being cleaned up. If not, that's an unknown bug.

Klemen2: I added `dbg!(doc.nodes.len());`  in the create_anonymous_block after create_node and the length keeps increasing just by hovering over in the dnd example without them having any hover attributes so I guess this is a unkown bug then, will add make an issue then.

Klemen2: removing "Files" and "Text" titles from the dnd example fixes this double enter issue, lmk if they trigger correctly, I think they do

Klemen2: double dragenter when dragging from the top down is correct as it hits "Files" node and since all dnd events are propagating the behaviour I believe is correct after #544 is closed with your fix

Klemen2: after the anonymous nodes fixes dragging drag me 2 times straight down to the green square produces exact same events as firefox so I'm marking this as resolved
drop events

nicoburns:
@Klemen2 AI-review has brought up quite a few issues:

----

## Blocking

1. **`DomEventData` / `DomEventKind` discriminant misalignment breaks the handler-count fast path.** Registration counts by `DomEventKind` discriminant (`mutation_writer.rs`: `event_handler_counts[kind.discriminant()]`), but dispatch reads `event.data.discriminant()` — `DomEventData`'s index (`dioxus_document.rs` `handle_event`). The new variants were added in different orders:
   - `DomEventKind`: `DragStart, Drag, DragEnter, DragOver, DragLeave, Drop, DragEnd`
   - `DomEventData`: `DragStart, Drag, DragEnd, DragEnter, DragOver, DragLeave, Drop`

   So `ondrop` reads the `ondragend` count, `ondragend` reads `ondragenter`, `ondragenter` reads `ondragover`, etc. A handler registered alone (e.g. only `ondrop`) is silently never invoked. The example masks this because every zone registers all four events. Easiest robust fix: index by `event.data.kind().discriminant()` (or reorder `DomEventData` to match `DomEventKind`).

2. **`DragOver` default action has `drop_effect` backwards.** `handle_dom_event` sets `drop_effect = Copy` in the default action (`blitz-dom/src/events/mod.rs`), but `run_default_action` only runs when the event was *not* canceled. Per spec a node isn't a drop target unless it cancels `dragover`, so as written every element that doesn't `preventDefault` advertises "copy" (droppable cursor everywhere), while elements that do `preventDefault` keep the previous value (initially `None`). Should be inverted: default `none`, set the effect when the event is canceled.

3. **`drop` is swallowed when no types can be fetched.** In `WindowEvent::DragDropped`, if `serial_ids` ends up empty (zero advertised types, or every `fetch_data_transfer` errors), `pending_data_transfers` is stored but no `DataTransferReceived` ever arrives, so `UiEvent::Drop` is never dispatched. Dispatch the drop immediately when `serial_ids.is_empty()`.

4. **`INTERNAL_DROP_THRESHOLD` heuristic is fragile.** `internal_drag` records the last `DragLeft`; an `OutgoingDragDropped` within 50ms is dispatched as an in-window `UiEvent::Drop`. A genuine drop just outside the window within 50ms of leaving fires `drop` on the last-hovered node — delivering the drag's data to a handler that shouldn't get it. At minimum this needs a comment naming the platform quirk it works around; ideally decide on actual drop position/geometry rather than timing.

5. **No in-progress guard for internal drags.** `doc.drag_mode = DragMode::None` inside the `drag_mode == DragMode::None` branch in `pointer.rs` is a no-op. Until the async `UiEvent::DragStart` clears `mousedown_node_id`, further pointermoves re-enter the branch and call `set_datatransfer` → `start_drag` again. Needs a `DragMode::Dnd`-style marker.

6. **Click + selection-clear fire after a completed drag.** Since `drag_mode` stays `None`, `do_click` in `handle_pointerup` is true — after a drop a `click` dispatches on the target and the added `doc.clear_text_selection()` wipes the selection. Browsers suppress click after a drag, and keep the source selection during a text drag.

7. **`start_internal_drag` dispatches `DragStart` before `event_loop.start_drag`.** If `start_drag` errors, the DOM got `dragstart` (with `id: 0`) but `dragend` never arrives — an unbalanced session. Assign `event.id` before dispatching and bail to `OutgoingCancel` on failure.

8. **Sub-document (iframe) drags will target the wrong node or panic.** `source.get_node_id()` is interpreted in the root document's node space — `handle_dom_event` indexes `doc.nodes[target]` directly — and `map_dom_event_to_ui_event` maps `DragStart`/`Drag`/`DragEnd`/`DragEnter`/`DragLeave` to `None`, so nothing forwards them. A drag sourced inside a sub-document hits a wrong node or an out-of-range index. Even if iframe DnD is out of scope for this PR, the panic case is worth a guard.

9. **`draggable` handling** (`pointer.rs`): only the hit element is checked — browsers use the nearest ancestor with `draggable=true`, so `div[draggable] > span` won't drag when grabbing the span. Also `v != "false"` enables dragging for invalid values like `draggable=""` — spec says only `"true"` enables it. `is_text_selected_at_position(x, y)` uses the current pointer position rather than `mousedown_position`, so pressing next to a selection and drifting onto it can start a text drag. And `draggable | auto_text_drag` should be `||`.

## Non-blocking

10. **`TextSelection::is_selected` orders endpoints by `(Option<NodeId>, sibling_index, offset)`** — node-id order isn't document order once the DOM mutates (ids are creation-order), and the scheme mixes `(parent, idx)` anonymous keys with regular node keys. Works for typical same-root selections but fragile; a document-order comparison would be more correct.

11. **`WinitDataTransferItems::get_data`/`types` require `type_().hint()`** — externally dropped items with a custom MIME type (no `TypeHint` mapping) are unreachable via `getData` and invisible in `types`, even though the OS delivered the data.

12. **`Box::leak` per unique MIME subtype** in `audio_mime_to_static_ext`/`image_mime_to_static_ext` — small unbounded leak driven by external MIME strings. A static match table or interning would avoid it.

13. **`set_dnd_action` is one event behind:** it reads `drop_effect` inside `update_active_drag` before `UiEvent::DragOver` is dispatched, so the OS sees the previous move's effect. Call it after `handle_ui_event` returns.

14. **Store-mode nits:** `readable()` set on `Drop` persists into `dragend` (spec says Protected there); `set_drop_effect` has no mode check. `WinitBlitzFile` does blocking `std::fs` inside async futures (blocks the UI executor for big drops) and returns `Some("unknown")` as content-type when the hint has no mime mapping — better to return `None` or infer from extension. `DragLeft` doesn't clear `doc.hover_node_id`, leaving stale `:hover`. The `url` dep added to blitz-shell appears unused.

Klemen2 (commented):
I'm not entirely sure what it means by 5. and 8., and I've added a todo for nearest ancestor of 9.
Edit: I haven't looked into 6 yet I kinda missed it while fixing the issues, it is technically also out of scope since internal event's weren't really planned for this pr and they still need some work

packages/blitz-shell/src/dnd.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        Self { data, types }
+    }
+}
+
+impl DataTransfer for WinitDataTransfer {
+    fn for_each_available_type<'this>(
+        &'this self,
+        func: &'_ mut dyn FnMut(&'this dyn TransferType) -> ControlFlow<()>,
+    ) {
```
Klemen2, line 169: this is what is causing me issues to resolve 7. in a better way due to lifetimes, see comment bellow

packages/blitz-shell/src/window.rs
```diff
@@ -582,7 +625,67 @@ impl<Rend: WindowRenderer> View<Rend> {
...
+            button: MouseEventButton::Main,
+            buttons: self.buttons,
+            data_transfer: data_transfer.clone(),
+            source: node.into(),
+        };
+
+        // this is done this way so that implementation of for_each_available_type
+        // for WinitDataTransfer actually works for outgoing drag events without
+        // issues with borrow checker or unsafe code, otherwise they will not have
+        // any data types within them
```
Klemen2, line 648: I haven't found a way to resolve 7. in a better way than to just send dragend if it fails

PR #498, opened 2026-07-09
Implement element_coordinates for wheel

Klemen2:
Continuation of #476 
Also handle secondary pointer/button out/leave/over/enter correctly

### DIFF
--- a/packages/blitz-dom/src/events/driver.rs
+++ b/packages/blitz-dom/src/events/driver.rs
@@ -1,6 +1,7 @@
 use crate::Document;
 use blitz_traits::events::{
-    BlitzPointerEvent, BlitzPointerId, DomEvent, DomEventData, EventState, UiEvent,
+    BlitzPointerEvent, BlitzPointerId, DomEvent, DomEventData, EventState, Point, PointerCoords,
+    UiEvent,
 };
 use std::collections::VecDeque;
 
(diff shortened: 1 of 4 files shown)


Klemen2:
should this be handeled in run_handler_event instead?

nicoburns:
Hmm... would it possible to put "element coordinates" resolving somewhere so that:

- It is one place (per event type)
- It doesn't require an additional clone of the event

?

nicoburns (approved):

nicoburns:
This needs to be rebase on top of main to fix the clippy lints

Klemen2:
ah, wrong way will fix this

Klemen2:
CI unrelated - no space left :p

packages/blitz-dom/src/events/driver.rs
```diff
@@ -328,6 +336,36 @@ impl<'doc, Handler: EventHandler> EventDriver<'doc, Handler> {
...
+            | DomEventData::PointerOver(data)
+            | DomEventData::PointerOut(data)
+            | DomEventData::MouseMove(data)
+            | DomEventData::MouseDown(data)
+            | DomEventData::MouseUp(data)
+            | DomEventData::MouseEnter(data)
+            | DomEventData::MouseLeave(data)
+            | DomEventData::MouseOver(data)
+            | DomEventData::MouseOut(data)
+            | DomEventData::TouchStart(data)
```
Klemen2, line 355: uf, I just rebased dnd that also needs these element coords and I noticed that DomEventData::TouchMove(data) is missing here :(
can you include it in some touch related pr so that I don't have to create a new one

nicoburns: Fixed in 406a571491523d69d52c396f240b14edd1a1974f

PR #499, opened 2026-07-09
Fix event bubbling

### LINKED ISSUE #496: Incorrect bubbling of events
The issue is caused by the fact that both blitz and vdom bubbles the event 

Expected result: 3 2 1
Blitz: 3 2 1 2 1 1

**Reference code**
https://github.com/DioxusLabs/blitz/blob/b21a161f3f930e5577a4693f8839e3e6006fa0f7/packages/dioxus-native-dom/src/dioxus_document.rs#L342-L344
https://github.com/DioxusLabs/blitz/blob/b21a161f3f930e5577a4693f8839e3e6006fa0f7/packages/blitz-dom/src/events/driver.rs#L324-L329

**Examples**
```rust

fn main() {
    dioxus_native::launch(app);
}

const CSS: &str = r#"
    div {
        border: 1px solid black;
        padding: 10px;
    }
 "#;

fn app() -> Element {
    rsx! {
        style {
            {CSS}
        }
        div {
            onclick: move |_| {
                dbg!("1");
            },
            "1"
            div {
                onclick: move |_| {
                    dbg!("2");
                },
                "2"
                div {
                    onclick: move |_| {
                        dbg!("3");
                    },
                    "3"
                }
            }
        }
    }
}
```

```html
<style>
    div {
        border: 1px solid black;
        padding: 10px;
    }
</style>
<div onclick="console.log('1')">
    1
    <div onclick="console.log('2')">
        2
        <div onclick="console.log('3')">3</div>
    </div>
</div>
```

nicoburns:
Fixes https://github.com/DioxusLabs/blitz/issues/496

Previously, everything was being bubbled n! times instead of n times as both dioxus-core and dioxus-native-dom were implementing bubbling.

@Klemen2 Does this fix the issue for you?

### DIFF
--- a/packages/dioxus-native-dom/src/dioxus_document.rs
+++ b/packages/dioxus-native-dom/src/dioxus_document.rs
@@ -330,27 +330,26 @@ impl EventHandler for DioxusEventHandler<'_> {
             return;
         };
 
-        for &node_id in chain {
-            // Get dioxus vdom id for node
-            let dioxus_id = doc.inner().get_node(node_id).and_then(get_dioxus_id);
-            let Some(id) = dioxus_id else {
-                continue;
-            };
+        // Get dioxus vdom id for node
+        let dioxus_id = chain
+            .iter()
+            .find_map(|node_id| doc.inner().get_node(*node_id).and_then(get_dioxus_id));
+        let Some(id) = dioxus_id else {
+            return;
+        };
 
-            // Handle event in vdom
-            let dx_event = Event::new(event_data.clone(), event.bubbles);
-            self.vdom
-                .runtime()
-                .handle_event(event.name(), dx_event.clone(), id);
+        // Handle event in vdom
+        let dx_event = Event::new(event_data.clone(), event.bubbles);
+        self.vdom
+            .runtime()
+            .handle_event(event.name(), dx_event.clone(), id);
 
-            // Update event state
-            if !dx_event.default_action_enabled() {
-                event_state.prevent_default();
-            }
-            if !dx_event.propagates() {
-                event_state.stop_propagation();
-                break;
-            }
+        // Update event state
+        if !dx_event.default_action_enabled() {
+            event_state.prevent_default();
+        }
+        if !dx_event.propagates() {
+            event_state.stop_propagation();
         }
     }
 }


Klemen2:
hmm, this now doesn't bubble at all

nicoburns:
Ack. Well we can go the other way. But that is going to break some work-in-progress code I have for adding events handlers to html/body

Klemen2:
the problem with this is that it's checking if that created event dx_event propagates and then breaks based on that, but removing that makes all propagate even the ones you add e.stop_propagation(); / are none propagating

nicoburns:
> the problem with this is that it's checking if that created event dx_event propagates and then breaks based on that, but removing that makes all propagate even the ones you add e.stop_propagation(); / are none propagating

Yep! I can see the problem. Unfortunately I don't think there's a full fix without modifying dioxus-core (except to let Dioxus handle the bubbling and not do bubbling in Blitz).

Klemen2:
welp, it's not that big of an issue so if you think that it would be less painful to wait on this then you can, it's just something that I noticed while trying out the drop events since they are all propagating by default

Klemen2:
also a weird thing that I noticed when you dbg a node_id in that loop is that with this when you click on a parent element e.g on 1 or 2 on that example in #496, the first node_id that gets skipped keeps increasing which is interesting (probably unrelated to this change)

nicoburns:
@Klemen2 Ok, I've implemented the fix the other way. Does this fix your issue?

Klemen2:
only for most inner child e.g 3 in that example, 2 and 1 are partially unclickable now
e.g clicking on the numbers now doesn't trigger anything, while on the padding it does, probably related to the comment above

nicoburns:
How about now?

Klemen2:
yep, this does now works, those shadow nodes should probably get noted somewhere, I think thats the same issue that I'm also experiencing when the ram keeps increasing, will make an issue with example later for that, I don't think thats the blitz problem

nicoburns:
> yep, this does now works,

Great :)

> those shadow nodes should probably get noted somewhere, I think thats the same issue that I'm also experiencing when the ram keeps increasing, will make an issue with example later for that, I don't think thats the blitz problem

Not sure I understand this yet. I will await the issue!

PR #506, opened 2026-07-11
Switch to struct mutation for renderer options

varappdev:
## Summary
This pull request changes how Renderer Options are constructed as DioxusLabs/anyrender#70 will mark them as `#[non_exhaustive]`.

We now construct the default options then modify them to our desired options.

This pull request has no breaking changes so it should be able to be accepted regardless of if DioxusLabs/anyrender#70 gets accepted.

## Key Changes
- `dioxus-native`: Change from constructing with struct notation and `::default()` to making the `::default()` then modifying it.
- `examples/wgpu_texture`: Change from constructing with struct notation and `::default()` to making the `::default()` then modifying it.


nicoburns (approved):

nicoburns:
Not necessarily convinced that this is better, but it seems fine. Probably we should make the options structs in AnyRender implement builder-pattern methods.

nicoburns:
Ok, there's a clippy lint specifically against this. Let's just not do this.

varappdev:
That clippy lint only triggers when the struct is not marked as non-exhaustive but I'll go ahead and implement the builder pattern in the other pull request.

PR #511, opened 2026-07-13
Add svg-text feature so consumers can opt out of usvg's text-shaping stack

koua-kigo:
## Motivation

`usvg`'s default features include `text` (SVG `<text>` shaping), which pulls in `rustybuzz` — recently flagged unmaintained by [RUSTSEC-2026-0206](https://rustsec.org/advisories/RUSTSEC-2026-0206) (maintainer announcement: harfbuzz/rustybuzz#166), with no fixed version available: `usvg` 0.47 still depends on it, and its suggested replacement (`harfrust`) hasn't been adopted upstream yet.

Consumers that never render `<text>` elements inside SVG documents — e.g. server-side HTML-to-image rendering where all text is HTML (shaped by parley) and SVGs are shape-only icons/logos/barcodes — currently can't drop that stack: Cargo feature unification is additive, and blitz-dom/blitz-paint's `svg` defaults force-enable `usvg/text` for every downstream. The advisory then fails `cargo deny`/`cargo audit` gates in those projects with no recourse except forking.

## Change

Adds an `svg-text` feature to `blitz-dom` and `blitz-paint`:

- **Included in both crates' `default` sets — default behavior is completely unchanged.**
- `blitz-dom`: `svg-text = ["svg", "usvg/text", "usvg/system-fonts", "usvg/memmap-fonts"]`; the workspace `usvg` dep drops default features; the `FONT_DB` glue and the `usvg::Options` font wiring are gated on `svg-text`.
- `blitz-paint`: pass-through `svg-text = ["svg", "blitz-dom/svg-text", "anyrender_svg/text"]`; the workspace `anyrender_svg` dep drops default features but keeps the four `image_format_*` features so raster images inside SVGs keep decoding.

With `default-features = false, features = ["svg"]`, usvg is built without its text stack: `rustybuzz`, `fontdb`, `fontconfig-parser`, and the `unicode-*` shaping crates leave the dependency tree; `<text>` elements inside SVG documents are not rendered (everything else — paths, shapes, gradients, embedded raster images — is unaffected).

## Testing

- `cargo check -p blitz-dom -p blitz-paint` (defaults, svg-text on) — compiles, lockfile unchanged in content.
- `cargo check -p blitz-dom -p blitz-paint --no-default-features --features svg` (opt-out) — compiles; `rustybuzz`/`fontdb` absent from the resolved graph.

Happy to adjust naming (`svg-text` vs `svg_text`), feature composition, or scope if you'd prefer a different shape.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,7 +112,7 @@ anyrender_skia = { version = "0.9.0" }
 anyrender_vello = { version = "0.12.0" }
 anyrender_vello_cpu = { version = "0.14.0" }
 anyrender_vello_hybrid = { version = "0.8.0" }
-anyrender_svg = { version = "0.12.0" }
+anyrender_svg = { version = "0.12.0", default-features = false, features = ["image_format_png", "image_format_gif", "image_format_jpeg", "image_format_webp"] }
 wgpu_context = { version = "0.7.0" }
 
 # Linebender + Fontations + WGPU + SVG
@@ -121,7 +121,7 @@ linebender_resource_handle = "0.1"
 peniko = "0.6.0"
 kurbo = "0.13.1"
 wgpu = "29"
-usvg = "0.46"
+usvg = { version = "0.46", default-features = false }
 
 # Windowing & Input
 raw-window-handle = "0.6.0"
--- a/packages/blitz-dom/Cargo.toml
+++ b/packages/blitz-dom/Cargo.toml
@@ -13,6 +13,7 @@ rust-version.workspace = true
 [features]
 default = [
     "svg",
+    "svg-text",
     "woff",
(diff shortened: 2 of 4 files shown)


nicoburns:
Hmm... we probably ought to finish up https://github.com/linebender/resvg/pull/922 instead. In the meantime I would recommend ignoring the RUSTSEC advisory. Note: there is no actual vulnerability here. Just lack of maintenance.

koua-kigo:
Thanks for the quick guidance — that makes sense. We'll ignore the advisory on our side and watch linebender/resvg#922. Closing this in favor of that work; happy to revive it if an opt-out feature ever becomes useful independently.

PR #550, opened 2026-07-28
Map the rest of HTML's semantic elements to AccessKit roles

marshallshelly:
`build_accessibility_node` maps nine elements and carries a `// TODO match more roles`, so everything else arrives as `Role::Unknown` — including most of what assistive technology actually navigates by.

On a small but fully semantic page (header, nav, main, a list, a table, a labelled input), **21 nodes came back `unknown`**:

```
window
  header
    heading
    unknown          <- <nav>
      unknown        <- <a>
  unknown            <- <main>
    unknown          <- <ul>
      unknown        <- <li>
    unknown          <- <table>
      unknown        <- <tr>
        unknown      <- <th>
    unknown          <- <label>
```

A screen reader gets no landmarks, no list structure, no table structure and no labels.

## What this adds

Mappings from [HTML-AAM](https://www.w3.org/TR/html-aam-1.0/):

| Group | Elements |
| --- | --- |
| Landmarks | `nav` `main` `aside` `footer` `article` `search` |
| Grouping | `ul`/`ol` `li` `dl`/`dt`/`dd` `fieldset` `form` `dialog` `blockquote` `figure` `figcaption` `hr` |
| Tables | `table` `thead`/`tbody`/`tfoot` `tr` `td` `th` |
| Interactive | `a` `label` `legend` `select` `option` `textarea` `progress` `meter` `output` `summary` |
| Inline | `code` `em` `strong` `mark` `time` `img` `iframe` |

`<input>` gains `button`/`submit`/`reset`, `color`, `date`, `datetime-local`, `email`, `password`, `radio`, `range`, `search`, `tel`, `time`.

Three mappings are conditional, per spec:

- `<a>` is a `Link` only with an `href`; without one it is a `GenericContainer`.
- `<th>` is a `RowHeader` when `scope` is `row`/`rowgroup`, `ColumnHeader` otherwise.
- `<select>` is a `ListBox` with `multiple`, `ComboBox` otherwise.

Existing mappings are untouched, and `previously_mapped_roles_are_unchanged` pins that.

## Testing

`packages/blitz-html/tests/accessibility_roles.rs` — 6 tests. The last one asserts that on a semantic page the only unmapped elements are `html`, `head` and `body`, and prints the offending tags on failure rather than a bare count.

The tests need `blitz-dom`s `accessibility` feature, which `blitz-html` did not expose, so it gains an `accessibility` feature forwarding to it:

```bash
cargo test -p blitz-html --features accessibility
```

[...]

PR #549, opened 2026-07-28
Stop `position: fixed` resolving against a positioned ancestor

marshallshelly:
Taffy has no `Fixed` position, so `stylo_taffy` maps it to `Absolute` ([`convert.rs:231`](https://github.com/DioxusLabs/blitz/blob/main/packages/stylo_taffy/src/convert.rs#L231)). An absolutely positioned node resolves its insets against its containing block, so a fixed node laid out in place resolves against its nearest positioned ancestor instead of the viewport.

## Repro

```html
<div style="position:relative;left:200px;top:150px;width:300px;height:200px">
  <div id="a" style="position:fixed;top:10px;left:10px;width:120px;height:40px"></div>
  <div id="b" style="position:fixed;inset:0"></div>
</div>
```

| | before | after | browser |
| --- | --- | --- | --- |
| `#a` | `210,160` | `10,10` | `10,10` |
| `#b` position | `200,150` | `0,0` | `0,0` |
| `#b` width | `300` | viewport width | viewport width |
| `#b` height | `200` | root element height | viewport height |

With no positioned ancestor the result was already correct, because `absolute` and `fixed` then share a containing block. That is probably why this went unnoticed.

## Approach

Fixed nodes are reparented onto the root element in a new `hoist_fixed_position_nodes`, called between `resolve_layout_children` and `flush_styles_to_layout`. Since `flush_styles_to_layout` derives `paint_children` from `layout_children`, painting and hit testing follow the reparent with no further work.

The hoist runs on every pass and is idempotent, so it holds across both branches of `resolve_layout_children` — the one that rebuilds `layout_children` and the one that preserves them. There is a test for that.

A transformed ancestor is the containing block for its fixed descendants per [css-transforms-1](https://drafts.csswg.org/css-transforms-1/#propdef-transform), so those are skipped. `filter`, `backdrop-filter`, `will-change`, `contain` and `perspective` also establish one; those are marked TODO rather than handled.

## What this does not fix

The root element takes its height from its content, whereas the initial containing block is always viewport-sized regardless of document height. So `inset: 0` still sizes against the document. Reparenting alone cannot express this — it needs an ICB node distinct from the root element, which felt like a larger design question than this PR should decide. `opposite_insets_should_size_against_the_viewport` is `#[ignore]`d with the browser-correct expectation, ready to be unignored if you add one.

Happy to take that on in a follow-up if you have a preferred shape for it.

## Testing

[...]

PR #551, opened 2026-07-28
Give the focus outline a style, so it is actually painted

marshallshelly:
The UA stylesheet already has the right rule:

```css
input:focus,
textarea:focus {
    outline: 2px #4D90FE;
}
```

but the `outline` shorthand omits the style keyword, so [`outline-style` resets to `none`](https://drafts.csswg.org/css-ui/#outline) and nothing is drawn. **Keyboard users get no focus indicator on any input or textarea.**

Adding `solid` is the whole fix.

## This is not an outline-painting bug

Outline painting works. Same page, same engine:

| CSS | Result |
| --- | --- |
| `outline: 2px #4D90FE` (the current UA rule) | nothing painted |
| `outline: 2px solid #4D90FE` | ring painted correctly |

So the rule was simply inert.

## Testing

`packages/blitz-html/tests/focus_outline.rs` asserts an idle field resolves to no outline and a focused one resolves to something paintable, for both `<input>` and `<textarea>`. Reverting the keyword makes both tests fail, so it cannot silently regress.

Found while comparing a React and Tailwind app rendered by Blitz against the same app in Chrome: everything matched except the focus ring.

Independent of #549 and #550.

---

#### Devin Review
| Status | Commit |
|---|---|
| ⚪ Not started | — |

> [Run Devin Review](https://dioxus.staging.devinenterprise.com/review/DioxusLabs/blitz/pull/551?analyze=true)

> 💡 [Connect your GitHub account](https://dioxus.staging.devinenterprise.com/settings/profile#linked-accounts) to enable automatic code reviews.

<a href="https://dioxus.staging.devinenterprise.com/review/DioxusLabs/blitz/pull/551" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-staging-dark.svg?v=1">
    Open in Devin Review (Staging)
  </picture>
</a>

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/30400697657).</sub>

### DIFF
--- a/packages/blitz-dom/assets/default.css
+++ b/packages/blitz-dom/assets/default.css
@@ -74,7 +74,7 @@ textarea {
 
 input:focus,
 textarea:focus {
-    outline: 2px #4D90FE;
+    outline: 2px solid #4D90FE;
 }
 
 button,


nicoburns (changes requested):
Fix looks good, but can we remove the test. That doesn't look useful.

marshallshelly:
Test removed — the fix is just the one-word CSS change now.

nicoburns (approved):

PR #552, opened 2026-07-28
Render the placeholder attribute on inputs and textareas

marshallshelly:
The `placeholder` attribute is ignored entirely today. It appears nowhere in `blitz-dom` or `blitz-paint`, so an empty field renders as a blank box with no hint of what belongs in it.

## Approach

`TextInputData` gains a second `PlainEditor` holding the placeholder, laid out with the same styles as the value.

Keeping it separate is the point. Seeding the main editor with the placeholder text would be far less code, but the placeholder would then be indistinguishable from a real value to form serialisation, to `editor.text()`, and to anything scripting `.value`.

Paint shows it while the editor is empty, dimmed via `push_layer` alpha rather than a separate colour, so it tracks the element `color`. Styling it directly needs `::placeholder` support, which this deliberately does not attempt.

## It also drops the single-space seed

An empty input was seeded with `" "`:

```rust
editor.set_text(element.attr(local_name!("value")).unwrap_or(" "));
```

That space gave the editor a line box, but it was indistinguishable from a real value. An untouched field reported `" "` rather than `""`, and after typing `hey` it reported `"hey "`. It also makes "is this field empty?" ambiguous, which the placeholder needs to answer.

Removing it is included here because the two are entangled; the full `blitz-dom` and `blitz-html` suites pass unchanged with it gone. Happy to split it out if you would rather review it separately.

## Testing

`packages/blitz-html/tests/input_placeholder.rs` covers: an empty input shows its placeholder and reports an empty value, a value hides it, no attribute means none, an empty attribute is ignored, and `<textarea>` works too.

Found while rendering a React and Tailwind app with Blitz and comparing it to Chrome. Independent of #549, #550 and #551.

---

#### Devin Review
| Status | Commit |
|---|---|
| ⚪ Not started | — |

> [Run Devin Review](https://dioxus.staging.devinenterprise.com/review/DioxusLabs/blitz/pull/552?analyze=true)

> 💡 [Connect your GitHub account](https://dioxus.staging.devinenterprise.com/settings/profile#linked-accounts) to enable automatic code reviews.

<a href="https://dioxus.staging.devinenterprise.com/review/DioxusLabs/blitz/pull/552" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-staging-dark.svg?v=1">
    Open in Devin Review (Staging)
  </picture>
</a>

## WPT results

**0** newly passing, **2** newly failing (net -2).

<details>
<summary>Full diff (2 changed tests)</summary>

```diff
- Pass => Fail css/css-shadow/slotted-placeholder.html
- Pass => Fail css/css-values/attr-pseudo-element-placeholder.html
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/32318197532).</sub>

### DIFF
--- a/packages/blitz-dom/assets/default.css
+++ b/packages/blitz-dom/assets/default.css
@@ -77,6 +77,10 @@ textarea:focus {
     outline: 2px #4D90FE;
 }
 
+::placeholder {
+    color: color-mix(in srgb, currentColor 54%, transparent);
+}
+
 button,
 input[type="submit"],
 input[type="reset"],
--- a/packages/blitz-dom/assets/default.css
+++ b/packages/blitz-dom/assets/default.css
@@ -79,6 +79,10 @@ textarea:focus {
     outline: 2px solid #4D90FE;
 }
 
+::placeholder {
+    color: color-mix(in srgb, currentColor 54%, transparent);
+}
+
 button,
 input[type="submit"],
 input[type="reset"],
--- a/packages/blitz-dom/src/layout/construct.rs
+++ b/packages/blitz-dom/src/layout/construct.rs
@@ -2,6 +2,7 @@ use blitz_traits::node_id::NodeId;
 use core::str;
 use std::sync::Arc;
 
+use color::{AlphaColor, Srgb};
 use markup5ever::{QualName, local_name, ns};
 use parley::{
     FontContext, InlineBox, InlineBoxKind, LayoutContext, StyleProperty, TreeBuilder,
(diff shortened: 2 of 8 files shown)


nicoburns (changes requested):
This should use a `parley::Layout` rather than a `parley::PlainEditor`: placeholder text is not editable. Ideally it should also resolve the styles for the `::placeholder` pseudo-element (it's common for stylesheets to override things like the text color for placeholder text).

marshallshelly:
Switched to `parley::Layout` and resolved `::placeholder` as a lazy pseudo-element.

One wrinkle worth flagging, because it shaped the patch: the colour can't ride along on the resolved style. `stroke_text` resolves a glyph run's colour by looking up the node its `TextBrush` points at, and that node is the input element — whose `color` is exactly what `::placeholder` exists to override. So a resolved pseudo style can't reach paint through the brush at all.

What I did instead: carry the colour on the placeholder next to its layout, give `stroke_text` an override parameter, and put the default in the UA sheet where a stylesheet can replace it:

```css
::placeholder { color: color-mix(in srgb, currentColor 54%, transparent); }
```

That also replaces the paint-time alpha layer the previous version used to dim the text.

Happy to switch to an anonymous style-carrier node instead if you'd rather keep the brush indirection uniform — that matches how `::before`/`::after` work, at the cost of managing the lifetime of a node that is never laid out.

marshallshelly:
Rebased onto main. The text painter changed a fair bit underneath this, so a couple of notes:

- The test moved to `tests/blitz-tests/tests/input_placeholder.rs`, since `packages/blitz-html/tests/` is gone.
- `stroke_text` now resolves a run's colour from the ancestor stack rather than the brush node, so `color_override` is an extra argument on the end. All seven tests still pass.

One question while you're looking: now that colour comes from per-node resolved styles, would you rather `::placeholder` go through that instead of an override argument? That would drop the parameter entirely, but it wasn't obvious to me whether a pseudo-element fits into the decoration stack cleanly. Happy to take it that way if you prefer.

PR #556, opened 2026-07-29
Update parley to 0.11

marshallshelly:
Compiles unchanged across the workspace, and the `blitz-dom` and `blitz-html` test suites pass without modification.

**One tradeoff to weigh, since it works against the note on the `skrifa` pin.** parley 0.11 depends on skrifa 0.43, while vello 0.9 still requires 0.42, so the tree ends up carrying both where `main` carries one:

```
skrifa 0.42.1   <- vello 0.9, vello_encoding 0.9
skrifa 0.43.2   <- parley 0.11
```

Bumping the `skrifa` pin alongside does not resolve it — vello holds 0.42 either way — so the duplicate stays until vello updates. Happy to close this if you would rather take parley and skrifa together when that lands.

For what it is worth, I checked the other pins that are behind, and none of them move cleanly yet:

| | | |
|---|---|---|
| `wgpu` 29 → 30 | duplicates | vello 0.9, wgpu_context 0.8 require ^29 |
| `usvg` 0.46 → 0.47 | duplicates | anyrender_svg 0.13 requires ^0.46 |
| `pollster` 0.4 → 1.0 | duplicates | anyrender_vello 0.13, rfd 0.17 require ^0.4 |
| `keyboard-types` 0.7 → 0.8 | already both | winit-core 0.31 wants 0.8, dioxus-html 0.7 wants 0.7 |

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31338666354).</sub>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2902,9 +2902,9 @@ dependencies = [
 
 [[package]]
 name = "fontique"
-version = "0.10.0"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "274fa4f0f0a926ae182c7c076c078cce8a38471d15e61a102a02cac984be9813"
+checksum = "1e04c4750a17111ebd77c3e0aea00476ce33f59235bc4d9e7f0aded5033ad3fc"
 dependencies = [
  "hashbrown 0.17.1",
  "linebender_resource_handle",
@@ -2914,7 +2914,7 @@ dependencies = [
  "objc2-core-text",
  "objc2-foundation 0.3.2",
  "parlance",
- "read-fonts 0.39.2",
+ "read-fonts 0.40.2",
  "roxmltree 0.21.1",
  "smallvec",
  "windows",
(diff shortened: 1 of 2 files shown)


nicoburns:
We're planning coordinated releases of Parley, Vello, and resvg with matching Skrifa soon. But can merge this now if you need it sooner, seeing as release is currently blocked by git dep on Stylo anyway.

marshallshelly:
No rush from my side — the coordinated release sounds much better than carrying the duplicate skrifa, so I'd rather this waited for that.

Nothing here needs parley 0.11 specifically. I opened it as currency after checking which pins were behind, and it was the only one that unified at all; the value was mostly in finding out that the rest are pinned by vello/anyrender_svg/dioxus-html rather than by anything blitz controls. Happy for you to close this and take parley, vello, resvg and skrifa together, or to leave it open as a reminder — whichever is less noise for you.

marshallshelly:
Closing this — main is on parley 0.11.1 now, so the coordinated release covered it.

PR #600, opened 2026-08-02
Draw position-inside list markers in the bullet font

marshallshelly:
`font_for_bullet_style` gives position-**outside** markers the bundled bullet font. The position-**inside** path in `build_inline_layout_into` pushes the marker straight into the inline builder using the element's own style, so it never asks for that font.

Bullet glyphs are not in a typical text font, so an inside marker falls through to system fallback — and can land on nothing.

## Where it shows

`details > summary`. The UA sheet styles it `list-style: disclosure-closed inside`, so every disclosure triangle takes the inside path. On a machine whose fallback chain has no font covering U+25B8, the triangle renders as a missing-glyph box instead — reproducible on macOS 15 with any page containing a plain `<details>`.

The glyph is not missing from the machine — naming Menlo, Arial Unicode MS or STIXGeneral explicitly renders `▸ ▾ ✓` fine. Fallback simply does not reach them, which is a separate question; the marker should not be depending on fallback in the first place when the bullet font is bundled for exactly this.

## The change

Push a style span requesting the same family the outside path asks for, around the char marker only. `Marker::String` markers — `decimal`, `lower-alpha` and friends — are ordinary text and keep the element's font, which is what you want.

The family string moves into one `BULLET_FONT_FAMILY` constant so the two paths cannot drift apart again.

## Checking it

Built against the published `blitz-dom` `0.3.0-beta.1` with this patch applied and rendered a page with `<details>` through [Kiln](https://github.com/marshallshelly/kiln): the missing-glyph box becomes a triangle. `cargo fmt --all --check` and `cargo clippy -p blitz-dom --all-features` are clean.

I have not added a test. Marker rendering lands in a Parley layout rather than the DOM tree, so asserting it means asserting on shaped glyph runs, and I could not see an existing seam for that — happy to add one if you can point me at the right place.

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/30767394396).</sub>

### DIFF
--- a/packages/blitz-dom/src/layout/construct.rs
+++ b/packages/blitz-dom/src/layout/construct.rs
@@ -29,7 +29,11 @@ use crate::{
     traversal::{iter_children, iter_children_and_pseudos},
 };
 
-use super::{damage::ALL_DAMAGE, list::collect_list_item_children, table::build_table_context};
+use super::{
+    damage::ALL_DAMAGE,
+    list::{BULLET_FONT_FAMILY, collect_list_item_children},
+    table::build_table_context,
+};
 
 const DUMMY_NAME: QualName = qual_name!("div", html);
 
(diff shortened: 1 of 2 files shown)


nicoburns (approved):
Makes sense!

Issue #595, opened 2026-08-01
Transforms lag significantly sooner than in webview


Klemen2:
When transforming large number of children blitz tend to lag at a way smaller amount than the webview
Platform tested: windows 11

Blitz release count: **200**

https://github.com/user-attachments/assets/d24c23e4-75d4-4ea2-bf84-57651724f822

Webview release count: **3000**

https://github.com/user-attachments/assets/3670349e-4145-43b8-8ec2-b3c47d9f8cdf

```rust
use std::collections::HashMap;

use dioxus_native::prelude::dioxus_elements::geometry::ClientPoint;
use dioxus_native::prelude::*;

const COUNT: usize = 200;

pub fn main() {
    dioxus_native::launch(app);
}

fn app() -> Element {
    let mut viewport = use_store(Viewport::default);
    let data = use_store(Data::default);
    let mut position = use_store(|| ClientPoint::zero());

    let onpointermove = move |evt: Event<PointerData>| {
        let coords = evt.client_coordinates();
        viewport.write().pan(&coords, &position.peek());
        position.set(coords);
    };

    let onwheel = move |e: Event<WheelData>| {
        let coords = e.client_coordinates();
        let delta = e.delta().strip_units().y as f32;

        viewport
            .write()
            .onscroll(coords.x as f32, coords.y as f32, delta);
    };
    rsx!(
        style {
            {CSS}
        }
        div {
            tabindex: "-1",
            position: "absolute",
            top: 0,
            left: 0,
            width: "100vw",
            height:"100vh",
            onpointermove,
            onwheel,

            div {
                border: "1px solid black",
                position: "absolute",
                top: 0,
                left: 0,
                transform_origin: "0 0",
                width: 0,
                height: 0,
                transform: "{viewport}",
                div {
                    for (id, item) in data.elements().iter() {
                        PositionedItem{
                            id: id,
                            data: item
                        }
                    }
                }
            }
        }
    )
}

#[component]
fn PositionedItem(id: usize, data: Store<PositinedElement>) -> Element {
    rsx!(
        div {
            z_index: id,
            pointer_events: "none",
            user_select: "none",
            position: "absolute",
            border: "1px solid black",
            left: format!("{}px", data.x()),
            top: format!("{}px", data.y()),
            width: format!("{}px", data.width()),
            height: format!("{}px", data.height()),
            div {
                "Placed item {id}"
            }
        }
    )
}

#[derive(Store, Clone, PartialEq)]
[...]

PR #601, opened 2026-08-03
Fix: Windows not redrawing when requested

Klemen2:
related to: #595

Tested on windows 10 and 11

Digging into windows documentation WM_PAINT, that winit is using to sent redraw requests, is only sent by the system or when another application  makes a request to paint a portion of an application's window (https://learn.microsoft.com/en-us/windows/win32/gdi/wm-paint) and that it will only be sent when there are no messages in the queue.

With this fix constantly updating transforms work, this is **5000** elements while recording being smoother than **200** before.

https://github.com/user-attachments/assets/1f783694-13fa-4042-8ef4-8d621163ded0

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31265485113).</sub>

### DIFF
--- a/packages/blitz-shell/src/application.rs
+++ b/packages/blitz-shell/src/application.rs
@@ -105,6 +105,12 @@ impl<Rend: WindowRenderer> BlitzApplication<Rend> {
                     window.apply_pending_resize_if_settled();
                 }
             }
+            #[cfg(target_os = "windows")]
+            BlitzShellEvent::RequestWindowRedraw { window_id } => {
+                if let Some(window) = self.windows.get_mut(&window_id) {
+                    window.request_redraw();
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


nicoburns:
Hmm... seems like this should be opt-in at least. This will presumably completely remove any notion of "frame pacing" (even to the limited extent that Winit supports it)? We should perhaps ask around other Rust GUI communities to see how they solve this, because it doesn't seem Blitz-specific.

Klemen2:
I created an issue on winit and it is technically a 2 years old bug, in the meantime I can put it behind a feature instead if that is preferred. Something like "windows-immediate-redraw" or something like that? and change it so that it is deduped automatically and then redrawn in the event loop before? after? event handeling
We could also limit tickrate somehow, but it would not fix this issue 100%

nicoburns:
Yeah, something like that makes sense I think

Klemen2:
I chose redrawing after new event to avoid another variable for double redraws as WindowEvent::RedrawRequested may be sent if message queue is empty instead

Klemen2:
I removed the feature flag and instead went with coalescing, I believe this is a better approach

PR #604, opened 2026-08-04
Clip scrollable_overflow to border box for overflow:hidden/clip

jerry4718:
When an element has overflow:hidden or overflow:clip on an axis, its scrollable_overflow now gets clipped to the border box on that axis. This prevents hit testing from matching children that are visually clipped, which caused incorrect cursor and hover state when a child with a transform escaped its parent's overflow:hidden clipping.

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/30903210500).</sub>

### DIFF
--- a/packages/blitz-dom/src/resolve.rs
+++ b/packages/blitz-dom/src/resolve.rs
@@ -181,6 +181,18 @@ impl BaseDocument {
             overflow = overflow.union(child_rect_in_self);
         }
 
+        if let Some(style) = self.nodes[node_id].primary_styles() {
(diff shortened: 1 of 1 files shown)


nicoburns:
I definitely want to fix this, but I'm not convinced this is the right place. I think the fix probably ought to go in the hit testing code. Because the scrollable_overflow is also used for actual scrolling (which applies to Overflow::Hidden nodes which are programmatically scrollable!)

jerry4718:
I reopened it at #612

PR #614, opened 2026-08-05
Winit upgrade

Klemen2:
## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/33909233393).</sub>

### DIFF
--- a/packages/blitz-shell/src/convert_events.rs
+++ b/packages/blitz-shell/src/convert_events.rs
@@ -41,6 +41,11 @@ pub(crate) fn winit_ime_to_blitz(event: Ime) -> BlitzImeEvent {
             before_bytes,
             after_bytes,
         },
+        _ => {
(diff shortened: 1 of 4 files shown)


packages/blitz-shell/src/convert_events.rs
```diff
@@ -41,6 +41,7 @@ pub(crate) fn winit_ime_to_blitz(event: Ime) -> BlitzImeEvent {
             before_bytes,
             after_bytes,
         },
+        _ => unreachable!("unsupported Ime variant: {event:?}"),
```
nicoburns: These should be ignored rather than panicking. Otherwise if a new variant gets added everyones's apps will start panicking.

Klemen2: Ok I will change it to tracing error then and make it return some default values that makes sense to me

PR #615, opened 2026-08-05
Support the :focus-within pseudo-class

maurerdietmar:
Selector matching for :focus-within was stubbed to false and no element state was maintained for it, so styles keyed on it never applied - a common pattern for form field focus rings, where the visible border lives on a wrapper around the input.

Track ElementState::FOCUS_WITHIN on the focused element and all of its DOM ancestors. The chain comes from a new BaseDocument::node_ancestors helper next to node_layout_ancestors. The layout chain is not usable here because focus pertains to the DOM tree: it skips display: contents ancestors, and its links do not exist before the first layout pass, which is when autofocus fires. compare_document_order already did the same walk through a private helper, which is now that public one. The specification defines the containment over the flat tree; without a shadow DOM implementation to flatten, that is the DOM tree.

The old and new chains are diffed the way hover's are, so a focus move only restyles the parts they do not share. The state follows tree changes around the focused node: it is cleared on refocus, clear_focus and removal, and transferred from the old ancestor chain to the new one in DocumentMutator::add_children_to_parent when a subtree containing the focus moves to a different parent - including a move out of the document, where the focus is cleared with it and only the old chain is unset.

The first commit is a pre-existing gap this feature runs into rather than causes: set_focus_to and clear_focus dirty selector state without asking for a frame, so a keyboard focus change - which touches neither layout nor content - leaves the old focus styling on screen until an unrelated event brings a redraw along. Focus changes driven by a mutation, autofocus in particular, ride along on the mutator's redraw. Happy to split it into its own PR if you would rather keep this one to the pseudo-class.

Tests: focus_within.rs covers ancestors and blur, a display: contents ancestor, focus set before the first layout pass, removal of the subtree holding the focus, a move within the document, and a move out of it. focus_redraw.rs counts redraw requests through a recording shell for focus, blur, a repeated focus and a Tab keypress.

### DIFF
--- a/packages/blitz-dom/src/document.rs
+++ b/packages/blitz-dom/src/document.rs
@@ -868,6 +868,16 @@ impl BaseDocument {
         if self.focus_node_id == Some(node_id) {
             let shell_provider = self.shell_provider.clone();
             self.nodes[node_id].blur(shell_provider);
+            // Nothing on the chain contains the focus anymore. Only the surviving
+            // ancestors need a snapshot for restyling: nodes in the removed subtree had
+            // theirs dropped already and may be freed entirely.
+            for ancestor in self.node_ancestors(node_id) {
+                if self.nodes[ancestor].flags.is_in_document() {
+                    self.snapshot_node_and(ancestor, |node| node.remove_focus_within());
+                } else {
+                    self.nodes[ancestor].remove_focus_within();
+                }
+            }
             self.focus_node_id = None;
         }
         if self.mousedown_node_id == Some(node_id) {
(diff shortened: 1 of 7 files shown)


nicoburns (changes requested):
This needs to use regular ancestors (`node.parent`) rather than layout ancestors (`node.layout_parent`) as focus pertains to the DOM elements tree rather than the layout tree.

This also needs to handle updating `:focus-within` when an element is moved within the tree.

maurerdietmar:
Thanks, both points are addressed in the force-push (76c17eb).

maurerdietmar:
The series grew a commit in front: moving the focus did not request a frame, which is also fixed now.

PR #619, opened 2026-08-07
Measure both axes so width-only measurements are safe to cache

maurerdietmar:
Asking a box only for its width, when its width is already known, gets a height of zero back: taffy's block layout short-circuits, because the caller said it did not want the height. The measurement cache keys on the known dimensions and treats the requested axis as irrelevant, so that answer is stored where a later height query for the same box finds it, and the box reports no content at all.

Table cells are laid out at the width of their column, which is what makes them ask for this. A row then falls back to whatever height is specified elsewhere in it: a data table row carrying a spacer cell of `height: 22px` came out 22 tall around a 30px cell, clipping the text against the bottom of every row, and a row with no specified height anywhere collapsed onto its padding.

Answer width-only requests from the cache when a full measurement is already there, and otherwise measure without storing the result. The height a later query needs is computed once either way, so this trades no work for the correct answer.

## WPT results

**22** newly passing, **8** newly failing (net +14).

<details>
<summary>Full diff (30 changed tests)</summary>

```diff
+ Fail => Pass css/CSS2/backgrounds/background-position-applies-to-013d.xht
- Pass => Fail css/CSS2/margin-padding-clear/margin-collapse-101.xht
- Pass => Fail css/CSS2/margin-padding-clear/margin-collapse-105.xht
- Pass => Fail css/CSS2/margin-padding-clear/margin-collapse-113.xht
- Pass => Fail css/CSS2/margin-padding-clear/margin-collapse-115.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b01.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b02.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b03.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b04.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b05.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b06.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b07.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b08.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b09.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b10.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b11.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003b12.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c01.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c02.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c03.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c04.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c05.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c06.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c07.xht
+ Fail => Pass css/CSS2/tables/fixed-table-layout-003c08.xht
- Pass => Fail css/css-flexbox/flex-aspect-ratio-img-column-011.html
- Pass => Fail css/css-flexbox/flex-aspect-ratio-img-row-007.html
- Pass => Fail css/css-flexbox/flex-aspect-ratio-img-row-017.html
- Pass => Fail css/css-flexbox/flexbox-min-width-auto-002b.html
+ Fail => Pass css/css-grid/layout-algorithm/grid-find-fr-size-gutters-001.html
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31247339875).</sub>

### DIFF
--- a/packages/blitz-dom/src/layout/mod.rs
+++ b/packages/blitz-dom/src/layout/mod.rs
@@ -13,10 +13,10 @@ use style::Atom;
 use style::values::computed::CSSPixelLength;
 use style::values::computed::length_percentage::CalcLengthPercentage;
 use taffy::{
-    BlockContext, CollapsibleMarginSet, FlexDirection, LayoutPartialTree, NodeId, ResolveOrZero,
-    RoundTree, Style, TraversePartialTree, TraverseTree, compute_block_layout,
-    compute_cached_layout, compute_flexbox_layout, compute_grid_layout, compute_leaf_layout,
-    prelude::*,
+    BlockContext, CollapsibleMarginSet, FlexDirection, LayoutPartialTree, NodeId, RequestedAxis,
+    ResolveOrZero, RoundTree, RunMode, Style, TraversePartialTree, TraverseTree,
+    compute_block_layout, compute_cached_layout, compute_flexbox_layout, compute_grid_layout,
+    compute_leaf_layout, prelude::*,
 };
 
 pub(crate) mod construct;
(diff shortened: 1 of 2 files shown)


maurerdietmar:
Reworked: instead of keeping width-only answers out of the cache, width-only  requests are upgraded to both axes so every cached entry is complete — the exclusion made every width probe re-lay-out its subtree, which we measured as a 33x inline-layout overhead on deeply nested pages. 

Tests unchanged.

maurerdietmar:
Closing this as superseded. #776 moved blitz to the taffy release that fixes the measure cache's axis handling  on taffy's side, which is the problem this PR worked around from the blitz side: a width-only measurement no longer poisons the cached layout. I re-ran the regression tests I had carried for it (a table cell with a definite width still contributing its content height to the row, among others) against current main (a980f31) without this patch, and they pass.

PR #622, opened 2026-08-07
fix: scale CSS transform translate by viewport scale factor

jerry4718:
CSS transform values (translate, transform-origin) are in CSS pixels and must be multiplied by the viewport scale factor (hidpi_scale * zoom) when rendering to device pixels. Previously only layout positions were scaled while transform matrices were not, causing misalignment between layout coordinates and transformed elements on hidpi screens.

Pass the viewport scale into resolve_2d_transform and multiply the translation components (translate property, transform function m41/m42, and transform-origin) by it. Scale factors and rotation angles are unitless and remain unchanged.

Test: transform_viewport_scale.rs verifies translate is scaled at hidpi_scale=2 while scale(0.5) is not.

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31186432999).</sub>

### DIFF
--- a/packages/blitz-dom/src/node/node.rs
+++ b/packages/blitz-dom/src/node/node.rs
@@ -326,7 +326,7 @@ impl Node {
                 Point2D::new(CSSPixelLength::new(0.0), CSSPixelLength::new(0.0)),
                 Size2D::new(CSSPixelLength::new(w), CSSPixelLength::new(h)),
             );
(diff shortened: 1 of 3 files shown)


nicoburns:
Thanks for the PR. AI review found issues with it, and generated #623 as an alternative which should fix your issues without introducing regressions. I've merged that instead.

PR #648, opened 2026-08-08
Fix panic when a selector deposits parent flags on the Document

latentharbor:
`Node::selector_flags()` is generated by `element_accessors!`, so it exists only on Element and AnonymousBlock nodes. Stylo's `apply_selector_flags` deposits the `for_parent()` half of a match on the raw parent node:

```rust
if let Some(parent) = self.parent_node() {
    parent.selector_flags().set(parent.selector_flags().get() | parent_flags);
}
```

For the root `<html>` that parent is the Document node, where the field does not exist, and the accessor panics with ```selector_flags` is not available on this node kind`.

`for_parent()` covers `:first-child`, `:last-child`, `:only-child`, `:nth-child()`, `:nth-last-child()`, `~` and `+`, so the minimal reproduction needs no script, no network and no shadow DOM:

```html
<style>*:first-child { color: red }</style><p>hello</p>
```

`universal_accessors!` already carries a Document arm for `element_state` and `snapshot_handled` for exactly this reason; `selector_flags` was missed. This moves it to the same macro.

Found while running the Web Platform Tests against blitz-dom in [OpenKitesurf](https://github.com/latentharbor/OpenKitesurf), where it was breaking real pages — react.dev among them.

## WPT results

**8** newly passing, **3** newly failing (net +5), **2** other status changes.

<details>
<summary>Full diff (13 changed tests)</summary>

```diff
+ Crash => Pass css/CSS2/sec5/adjacent-002.xht
+ Crash => Pass css/CSS2/selectors/first-child-selector-003.xht
+ Crash => Pass css/CSS2/selectors/universal-selector-002.xht
- Pass => Fail css/css-align/self-alignment/self-align-safe-unsafe-flex-003.html
- Pass => Fail css/css-align/self-alignment/self-align-safe-unsafe-grid-003.html
- Pass => Fail css/css-text/text-align/text-align-match-parent-01.html
+ Fail => Pass css/css-text/text-align/text-align-match-parent-05.html
+ Fail => Pass css/css-writing-modes/direction-upright-002.html
+ Crash => Pass css/selectors/child-indexed-no-parent.html
! Crash => Fail css/selectors/invalidation/nth-child-in-shadow-root.html
! Crash => Fail css/selectors/invalidation/nth-last-child-in-shadow-root.html
+ Crash => Pass css/selectors/root-siblings.html
+ Crash => Pass css/selectors/selector-structural-pseudo-root.html
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31244960624).</sub>

### DIFF
--- a/packages/blitz-dom/src/node/element.rs
+++ b/packages/blitz-dom/src/node/element.rs
@@ -122,9 +122,12 @@ pub struct ElementData {
 /// The document node participates in layout and styling like an element, so it
 /// carries the same style/layout fields that were previously stored directly on
 /// [`Node`](super::Node).
-#[derive(Debug)]
 pub struct DocumentData {
     pub stylo_element_data: StyloData,
+    /// Selector flags deposited here by `apply_selector_flags` when a
+    /// `for_parent()` flag is applied while matching the root `<html>` element,
+    /// whose parent node is the document.
+    pub selector_flags: Cell<ElementSelectorFlags>,
     /// A clone of the document's shared style lock. Set when the owning
     /// [`Node`](super::Node) is constructed.
     pub guard: Option<SharedRwLock>,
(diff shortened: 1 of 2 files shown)


latentharbor:
CI hasn't run here, so for the record — local WPT run of `css/selectors`, upstream `d788124a` vs this branch: 6 crashes drop to 1. `child-indexed-no-parent`, `root-siblings` and `selector-structural-pseudo-root` go CRASH → PASS; the two `nth-*-child-in-shadow-root` invalidation tests go CRASH → FAIL (they now fail for unrelated shadow-DOM reasons instead of panicking). No other status in the suite changes. The one remaining crash (`not-default-ns-001`) is a different panic (`resolve.rs`, deferred inline-layout result applied to a node that is no longer an element) and is untouched by this PR.

nicoburns (approved):

nicoburns:
Makes sense. I wonder if we ought to just use `ElementData` in place of `DocumentData` rather than maintaining a separate struct.

nicoburns:
A couple of regressions here, but will prioritise fixing crashes.

PR #649, opened 2026-08-08
Carry needs_wrap into a display:contents child's layout children

latentharbor:
`collect_complex_layout_children` recurses into a `display: contents` child through `collect_layout_children`, which does not carry the enclosing container's `needs_wrap` policy. The text child is then pushed into a block container unwrapped, and taffy asks that Text node for its `style`, panicking with ```style` is not available on this node kind`.

Minimal reproduction, no script and no shadow DOM:

```html
<div><span style="display:contents">x</span><div>y</div></div>
```

`<slot>` is `display: contents` in the UA stylesheet, so this tends to surface as a shadow-DOM failure although it has nothing to do with shadow trees.

One behavioural note, measured rather than assumed: against the Web Platform Tests this clears a large share of the panics of this class but not all of them, and it changes `display-contents-suppression-dynamic-001` from PASS to FAIL. That test appears to have been passing because the subtree was suppressed by the bug rather than by the cascade, but I have not confirmed that reading and would welcome a second look.

Found while running WPT against blitz-dom in [OpenKitesurf](https://github.com/latentharbor/OpenKitesurf).

## WPT results

**8** newly passing, **1** newly failing (net +7), **24** other status changes.

<details>
<summary>Full diff (33 changed tests)</summary>

```diff
! Crash => Fail css/css-display/display-contents-before-after-001.html
+ Crash => Pass css/css-display/display-contents-button.html
! Crash => Fail css/css-display/display-contents-details.html
! Crash => Fail css/css-display/display-contents-dynamic-before-after-001.html
+ Crash => Pass css/css-display/display-contents-dynamic-flex-001-inline.html
+ Crash => Pass css/css-display/display-contents-dynamic-flex-001-none.html
! Crash => Fail css/css-display/display-contents-dynamic-flex-002-inline.html
! Crash => Fail css/css-display/display-contents-dynamic-flex-002-none.html
! Crash => Fail css/css-display/display-contents-dynamic-flex-003-inline.html
! Crash => Fail css/css-display/display-contents-dynamic-flex-003-none.html
! Crash => Fail css/css-display/display-contents-dynamic-inline-flex-001-inline.html
! Crash => Fail css/css-display/display-contents-dynamic-inline-flex-001-none.html
! Crash => Fail css/css-display/display-contents-dynamic-list-001-inline.html
! Crash => Fail css/css-display/display-contents-dynamic-list-001-none.html
! Crash => Fail css/css-display/display-contents-fieldset.html
+ Crash => Pass css/css-display/display-contents-flex-001.html
! Crash => Fail css/css-display/display-contents-flex-002.html
! Crash => Fail css/css-display/display-contents-flex-003.html
+ Crash => Pass css/css-display/display-contents-float-001.html
! Crash => Fail css/css-display/display-contents-inline-flex-001.html
! Crash => Fail css/css-display/display-contents-line-height.html
! Crash => Fail css/css-display/display-contents-list-001.html
+ Crash => Pass css/css-display/display-contents-oof-001.html
+ Crash => Pass css/css-display/display-contents-oof-002.html
! Crash => Fail css/css-display/display-contents-shadow-dom-1.html
- Pass => Fail css/css-display/display-contents-suppression-dynamic-001.html
! Crash => Fail css/css-display/display-contents-text-inherit.html
+ Crash => Pass css/css-display/display-contents-text-only-001.html
! Crash => Fail css/css-text/text-autospace/text-autospace-elements-002.html
! Crash => Fail css/css-text/text-autospace/text-autospace-elements-003.html
! Crash => Fail css/css-text/text-autospace/text-autospace-elements-004.html
! Crash => Fail css/css-text/text-autospace/text-autospace-elements-006.html
! Crash => Fail css/css-text/text-autospace/text-autospace-elements-007.html
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31244968318).</sub>

### DIFF
--- a/packages/blitz-dom/src/layout/construct.rs
+++ b/packages/blitz-dom/src/layout/construct.rs
@@ -338,11 +338,18 @@ fn classify_flow_children(
     }
 }
 
-pub(crate) fn collect_layout_children(
+/// Per-container preparation that must run whenever layout children are
+/// (re)collected, whichever collection strategy is used: resets construction
+/// flags, drops any stale inline layout, generates/updates pseudo-elements,
+/// sets up special elements (text inputs, checkboxes, svg) and numbers list
+/// items.
+///
+/// Returns `false` if the container was fully handled as a special element or
+/// has nothing to collect, in which case child collection must not proceed.
+fn prepare_layout_children_collection(
     doc: &mut BaseDocument,
     container_node_id: NodeId,
-    out: &mut LayoutChildren,
-) {
+) -> bool {
     // Reset construction flags
     // TODO: make incremental and only remove this if the element is no longer an inline root
     doc.nodes[container_node_id]
@@ -364,16 +371,16 @@ pub(crate) fn collect_layout_children(
                 .and_then(|el| el.attr(local_name!("type")));
             if tag_name == "textarea" {
                 create_text_editor(doc, container_node_id, true);
-                return;
+                return false;
             } else if matches!(
                 type_attr,
                 None | Some("text" | "password" | "email" | "number" | "search" | "tel" | "url")
             ) {
                 create_text_editor(doc, container_node_id, false);
-                return;
+                return false;
             } else if matches!(type_attr, Some("checkbox" | "radio")) {
                 create_checkbox_input(doc, container_node_id);
-                return;
+                return false;
             }
         }
 
@@ -414,7 +421,7 @@ pub(crate) fn collect_layout_children(
                     let _ = err;
                 }
             };
-            return;
+            return false;
         }
 
         //Only ol tags have start and reversed attributes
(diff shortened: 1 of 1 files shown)


nicoburns:
The issue you've identified seems valid, but I don't think the fix is quite right. Specifically, there is a bunch of code at the the top of `collect_layout_children` which now wouldn't run but needs to.

latentharbor:
Good catch, thanks — the direct recursion did skip everything at the top of `collect_layout_children` (construction-flag reset, stale inline-layout removal, `flush_pseudo_elements`, special-element setup, list numbering).

Pushed 93005ac1: that preamble is now extracted into `prepare_layout_children_collection`, which both `collect_layout_children` and the `display: contents` recursion in `collect_complex_layout_children` run before collecting. The recursion also consumes the construction damage on the contents element, matching the hoisting path. `collect_layout_children` itself is a pure extraction — no behaviour change on that path.

Measured effect (local WPT run of `css/css-display` + `css/css-text/text-autospace`, old fix vs new): `display-contents-suppression-dynamic-001` returns FAIL → PASS — so the one regression in the original version was exactly this skipped preamble (the dynamic test depends on the construction-flag reset), not cascade suppression as I'd guessed. Every other status is unchanged, including an identical crash list. Relative to `main` the PR should now be 8 newly passing, 0 newly failing.

nicoburns:
I think this has now been covered by #657 (apologies on delay and duplicating this work - pointing AI's at bug has resulted in a bunch of duplicate fixes: currently trying to burn way to down the open PRs to get back to a sensible state).

PR #650, opened 2026-08-08
Map the dir attribute with selectors the engine can parse

maurerdietmar:
The bidi rules in the default stylesheet came from Gecko and match through its internal pseudo-classes (:-moz-dir-attr-rtl and friends), which the servo selector parser does not know. It dropped every one of them, so dir="rtl" set no direction and a page could not switch to right-to-left layout at all, even though the layout side fully supports it.

Use the attribute-selector form from the HTML spec's bidi rendering section instead. dir=auto keeps its isolation but not the content-based direction, which needs bidi analysis no selector can express.

## WPT results

**3** newly passing, **2** newly failing (net +1).

<details>
<summary>Full diff (5 changed tests)</summary>

```diff
+ Fail => Pass css/css-align/self-alignment/self-align-safe-unsafe-flex-003.html
+ Fail => Pass css/css-align/self-alignment/self-align-safe-unsafe-grid-003.html
+ Fail => Pass css/css-text/text-align/text-align-match-parent-01.html
- Pass => Fail css/css-text/text-align/text-align-match-parent-05.html
- Pass => Fail css/css-writing-modes/direction-upright-002.html
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31249884360).</sub>


nicoburns:
Seems reasonable in the short term. Long-term we'll likely need full BiDi analysis and to bring back a similar system as Mozilla have.

PR #662, opened 2026-08-09
Map the HTML presentational attributes per spec

latentharbor:
Two commits reworking the legacy presentational-attribute mapping in `blitz-dom`.

**Map the HTML dimension attributes per spec**

The old parser was `parse::<f32>()` with any trailing unit peeled off, which

- rejected valid values like `200in` and `200 %` (trailing garbage is ignored per spec),
- accepted `-200` and `+200` (dimension values have no sign),
- read `20.25e2` as 2025px — the [rules for parsing dimension values](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-dimension-values) have no exponent, so that is 20.25px followed by junk.

It also fills in the missing parts of the element table: `width`/`height` on the embedded elements (`iframe`, `embed`, `video`, `object`, `img`, `marquee`, `input[type=image]`) and the table elements (`table`, `colgroup`, `col`, `tr`, `td`, `th`, with the ignore-zero rules for cells and table width), and `hspace`/`vspace` mapping to the margin pairs.

**Map border and the body margin attributes**

- `border` on `img`/`object`/`input[type=image]` follows the rules for parsing non-negative integers and maps to the four border widths plus `border-style: solid` on all four sides — width alone computes back to zero against the default `border-style: none`.
- `marginwidth`/`marginheight`/`topmargin`/`leftmargin` on `body` map to the margins the HTML standard lists for them. Deliberately no `rightmargin`/`bottommargin`: the spec does not define them, and WPT `body-margin-3a`/`3b` assert they are ignored.

## Validation

WPT via the in-tree runner:

- `html/rendering`: three tests go FAIL→PASS — `non-replaced-elements/tables/table-row-height.html`, `non-replaced-elements/the-hr-element-0/width.html`, `replaced-elements/attributes-for-embedded-content-and-images/img_border_percent.xhtml` — with no other status changes.
- `css/css-display` and `css/css-tables`: per-test statuses identical to main.

The body-margin and hspace/vspace mappings have no reftest signal in the runner (the WPT tests covering them are testharness.js-based, which the runner skips, and `body-margin-1`/`2` crash on an unrelated pre-existing iframe issue). I verified those with throwaway local reftests (e.g. `<body marginwidth=100 marginheight=30>` against `body { margin: 30px 100px }`, and `` against `margin: 10px 40px`): all fail on main and pass with this change.

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/31347480300).</sub>

### DIFF
--- a/packages/blitz-dom/src/stylo.rs
+++ b/packages/blitz-dom/src/stylo.rs
@@ -861,25 +861,79 @@ impl<'a> TElement for BlitzNode<'a> {
             None
         }
 
+        /// The HTML "rules for parsing dimension values".
+        ///
+        /// https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-dimension-values
+        ///
+        /// This is emphatically not `str::parse::<f32>()` with the unit peeled
+        /// off. The differences all show up in real attribute values:
+        ///
+        ///   * leading whitespace is skipped, so `"   00523   "` is 523px,
+        ///     but a sign is not a digit, so `"+200"` and `"-200"` are errors
+        ///     where `f32::from_str` happily accepts both;
+        ///   * anything after the number is ignored rather than fatal, so
+        ///     `"200in"`, `"200 %"` and `"200 abc"` are all 200px;
+        ///   * only a `%` *immediately* after the digits makes a percentage,
+        ///     which is why `"200%abc"` is 200% but `"200 %"` is 200px;
+        ///   * there is no exponent, so `"20.25e2"` is 20.25px and not 2025px
+        ///     -- `f32::from_str` gets this one wrong silently, which is worse
+        ///     than rejecting it;
+        ///   * a trailing `.` is allowed and consumes no digits, so `"200."`
+        ///     is 200px and `"200.%"` is 200%.
+        fn parse_dimension_attr(value: &str) -> Option<(f32, bool)> {
+            let bytes = value.as_bytes();
+            let mut i = 0;
+
+            // Skip ASCII whitespace.
+            while i < bytes.len() && matches!(bytes[i], b' ' | b'\t' | b'\n' | b'\x0C' | b'\r') {
+                i += 1;
+            }
+
+            // Must begin with a digit. A sign or a bare `.` is an error.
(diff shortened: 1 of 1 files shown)


nicoburns:
I like the concept here, but you should use the parsing functions from Stylo instead of implementing yourself where possible: https://docs.rs/stylo/latest/style/servo/attr/index.html

latentharbor:
Done in aa54fba2: `parse_size_attr` now goes through `style::servo::attr::parse_length` / `parse_nonzero_length` (the latter subsumes the hand-rolled ignoring-zero check), and the pixel-length attributes (`border`, the `body` margins) go through `parse_unsigned_integer`. Both hand-rolled parsers are deleted.

Two pre-existing parsers I deliberately left alone, since they predate this PR: the `bgcolor` hex parser (could move to `parse_legacy_color` as a follow-up — that would also pick up named colors and the legacy color algorithm) and the SVG `width`/`height` parser (those attributes take CSS `<length-percentage>` values, which the HTML microsyntax parsers don't cover).

Re-validated against WPT html/rendering: same three FAIL→PASS as before, no other status changes.

nicoburns:
Hmm... weird that this isn't changing any WPT tests

nicoburns:
But it does look correct to me.

nicoburns:
@latentharbor I'd be interested if you have an actual use case for the body margin and hspace/vspace attributes. Those don't seem very common?

latentharbor:
On the WPT numbers: CI only runs the `css/` suites — the three tests this PR flips are all in `html/rendering` (listed under Validation in the description).

Use case: I'm building an agent-oriented browser on top of blitz (OpenKitesurf — not released yet) that needs to render arbitrary real-world pages, including the legacy tail. `<body topmargin=0 leftmargin=0 marginwidth=0 marginheight=0>` is still standard boilerplate in HTML email, and hspace/vspace likewise on images in that kind of content. Once the parser was fixed, the remaining pixel-length rows of the hint table were a handful of `push_style` calls each — happy to drop those two families if you'd rather keep the surface smaller.

PR #760, opened 2026-08-20
Keep transforms in the 2d subset instead of dropping all 3d functions

ruifigueira:
`has_3d` tells you a 3d function appeared in the transform list, not that the resulting matrix needs 3d. The filter treats it as the latter, so `translate3d(x, y, 0)` gets dropped and the element paints with no transform. That one hurts, it's the standard centring trick.

`t.is_2d()` is the actual question, and it's the two-dimensional subset that the spec link on the next line down already points at. rotateX and perspective still get dropped, so the `TODO: support 3D transforms` above still stands.

Tests in `tests/blitz-tests/tests/transform_2d_subset.rs`. The two translate3d ones fail without the fix. The rotateX one passes either way, it's there so nobody later "fixes" 3d by flattening it.

No existing test caught this, by the way. `subpixel-transform-changes-004.html` in css-transforms looks like it should, but it uses translate3d on the reference side too, so dropping them all leaves both sides matching. Can send a reftest to WPT if you want one.

## WPT results

**5** newly passing, **2** newly failing (net +3).

<details>
<summary>Full diff (7 changed tests)</summary>

```diff
+ Fail => Pass css/css-transforms/css-transform-3d-rotate3d-Z-negative.html
+ Fail => Pass css/css-transforms/css-transform-3d-rotate3d-Z-positive.html
+ Fail => Pass css/css-transforms/css-transform-3d-rotateZ-negative.html
+ Fail => Pass css/css-transforms/css-transform-3d-rotateZ-positive.html
- Pass => Fail css/css-transforms/subpixel-transform-changes-004.html
+ Fail => Pass css/css-transforms/transform3d-matrix3d-001.html
- Pass => Fail css/css-viewport/zoom/transform-matrix-3d.html
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/32628672299).</sub>

### DIFF
--- a/packages/blitz-dom/src/stylo_to_kurbo.rs
+++ b/packages/blitz-dom/src/stylo_to_kurbo.rs
@@ -56,8 +56,17 @@ pub fn resolve_2d_transform(
             .transform
             .to_transform_3d_matrix(Some(&reference_box))
             .ok()
-            .filter(|(_t, has_3d)| !has_3d)
-            .map(|(t, _has_3d)| {
+            // The `has_3d` flag means a 3d function appeared in the list, not that the
+            // matrix needs 3d: `translate3d(x, y, 0)` sets it while resolving to a
+            // plain 2d translation.
+            //
+            // TODO: support 3D transforms. Outside a 3d rendering context the spec
+            // still applies the transform as a painting effect -- its own example
+            // paints `rotateY(50deg)` narrower -- so dropping one is wrong. `is_2d()`
+            // is closer than the flag, not the rule.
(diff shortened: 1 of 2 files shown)


nicoburns (approved):
Change definitely looks good. I don't always bother with tests for small things like this (especially as long-term it's not correct behaviour (we'll want to actually support 3d transforms!)), but seems fine.

Needs rebase on latest main to pick up fixes for Rust 1.98 clippy lints.

PR #761, opened 2026-08-21
Support the `role` attribute when setting the role in an `AccessKitNode`

hovinen:
Previously, the ARIA role of a node was determined only from the HTML element name. Normally the role attribute overrides the implicit role given by the element name; see https://www.w3.org/TR/html-aria/#rules-wd. Exceptions for "strong native semantics" are not treated by this PR.

This does not touch the existing mapping of element names to implicit ARIA roles -- that still needs to be completed.

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/32467606614).</sub>


nicoburns (approved):

nicoburns:
Looks good, but somehow this has conflicts.

hovinen:
> Looks good, but somehow this has conflicts.

Forgot to update from the main branch first, sorry. Fixed now.

PR #790, opened 2026-08-25
Remove cumulative paint layer limit

VladimirKras:
## Motivation

Dense scenes silently stop emitting requested clip and effect layers once cumulative painting crosses the nominal 1,024-layer limit, so later content is painted with the wrong layer stack and becomes visibly corrupted even when frame rate is healthy. The cap is enforced in backend-independent paint code, so it affects CPU and GPU renderers alike.

## Why the cap can be removed

The cap was introduced at 28 in [596d419](https://github.com/DioxusLabs/blitz/commit/596d419b27f3cef6ceb590223bef50c388cf15f4) as a workaround for [Vello #644](https://github.com/linebender/vello/issues/644), then raised to 1,024 in [c74b019](https://github.com/DioxusLabs/blitz/commit/c74b019f82830352725209b7ca711372a3e6dd30). Vello fixed the underlying empty-clip encoding bug in [Vello #651](https://github.com/linebender/vello/pull/651), merged in August 2024; Blitz now uses Vello 0.10.0, so retaining the old global workaround causes silent data loss without protecting current renderers from that historical bug.

## Changes

- Remove the cumulative layer cap and its now-unused accounting.

## Validation

- `cargo test -p blitz-paint --lib`
- `cargo clippy -p blitz-paint --no-deps -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

The workspace Clippy command currently stops on an unrelated existing `needless_return` warning at `packages/blitz-dom/src/mutator.rs:1218` with Rust 1.96.1; the affected crate passes Clippy with dependency linting disabled.

> [!NOTE]
> AI-assistance disclosure: this change and PR description were prepared with assistance from Codex/GPT-5 and reviewed by the submitter.

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/32809074670).</sub>


nicoburns (changes requested):
This seems reasonable, but can we ditch the test please.

Vello Classic and Vello Hybrid both still have issues with resource allocation, but this isn't really guarding against it effectively.

VladimirKras:
Sorry about the useless test. Dropped it.

PR #814, opened 2026-08-29
Replace hand-built JS DOM binding plumbing in `blitz-vibey-script` with typed class layers and a per-instance sized own-data registry

jerry4718:
## Summary

This rewrites the JS DOM class definitions in `blitz-vibey-script` onto a typed layer ("ExtendLayer") scheme: every DOM interface is a layer whose own data lives in a single per-instance sized-slot registry, prototypes are linked with ES class semantics, and own-block access is a plain Rust-side slot borrow plus a `TypeId` downcast. Everything is internal to `blitz-vibey-script`; the DOM/JS behavior surface is unchanged (covered by the existing `dom.rs` / `preact.rs` integration tests).

### Layers

Each interface is an `ExtendLayer` chained through a compile-time `Parent` type, with its own data behind an accessor:

- `Node` = `Extended<NodeLayer { node_id }>` (root layer); `CharacterData` / `Element` / `Document` extend `NodeLayer`
- `Event` = `Extended<EventLayer>`: configuration (`type`, `target`, `bubbles`, `cancelable`) and dispatch flags (`prevented`, `stopped`, `currentTarget` via `GcRefCell`) live in the own block
- `CSSStyleDeclaration` / `ComputedStyle` carry the styled node id

Prototypes are wired with `link_prototype` (`child.prototype.__proto__ = parent.prototype`, `child.constructor.__proto__ = parent.constructor`). `Node`, `Document`, `Element`, `CharacterData` and `Event` are registered classes, so `new Event(type, init)` is a working constructor, `instanceof Node/Element/CharacterData` answers truthfully through the linked prototype chain (`HTMLElement` aliases `Element`), and `on<event>` IDL properties live on the `Node` prototype.

Node wrappers are built from their layer chain via `from_chain!`, backed by the `RuntimeState::node_wrappers` identity cache.

### `OwnDataRegistry`

Each instance's native data slot holds an `OwnDataRegistry` (`Vec<GcRefCell<Option<Box<dyn OwnSlot>>>>`), sized at attach time by the leaf layer's compile-time `OwnBlock::DEPTH`; each layer addresses its slot with `OwnBlock::IDX`. The design is ported from [blitz-boa-demo](https://github.com/jerry4718/blitz-boa-demo).

- `with_own` / `with_own_mut` / `set_own_block` operate purely on the registry: slot borrows + `TypeId` downcasts, entirely on the Rust side. Wrapper construction fills the registry slots through `from_chain!`.
- GC safety: slots are `Box<dyn OwnSlot>` with `Trace` on the trait, so `JsValue`s inside layers (e.g. `EventLayer.target`) stay reachable. Slot access goes through `Option::as_deref` to reach `&dyn OwnSlot` directly — resolving `as_any_ref` on `&Box<dyn OwnSlot>` would hit the blanket impl on the box itself and break every downcast (the blanket impl also applies to the box, since `Box<dyn OwnSlot>: Any + Trace`)

## Tested

- `cargo test -p blitz-vibey-script`: 21/21 `dom.rs`, 2/2 `preact.rs` (real Preact render + todo interaction), doctests pass
- `cargo check --workspace` clean
- `cargo build -p browser --features javascript` builds

## WPT results

**1** newly passing, **15** newly failing (net -14), **23** other status changes.

<details>
<summary>Full diff (39 changed tests)</summary>

```diff
! Fail => Timeout css/css-anchor-position/anchor-position-inline-001.html
! Fail => Timeout css/css-anchor-position/anchor-position-inline-002.html
! Fail => Timeout css/css-anchor-position/anchor-position-inline-003.html
- Pass => Fail css/css-color/nested-color-mix-with-currentcolor.html
! Fail => Crash css/css-display/display-contents-dynamic-fieldset-legend-001.html
- Pass => Fail css/css-env/env-in-custom-properties.tentative.html
- Pass => Crash css/css-flexbox/dynamic-bsize-change.html
! Fail => Crash css/css-fonts/font-display/font-display-feature-policy-02.tentative.html
! Fail => Crash css/css-fonts/font-display/font-display.html
! Fail => Crash css/css-fonts/variations/font-parse-numeric-stretch-style-weight.html
- Pass => Crash css/css-grid/grid-definition/grid-change-intrinsic-size-with-auto-repeat-tracks-001.html
- Pass => Crash css/css-grid/grid-model/grid-layout-stale-002.html
! Fail => Crash css/css-images/image-orientation/image-orientation-none-cross-origin-svg.html
! Fail => Crash css/css-overflow/dynamic-visible-to-clip-001.html
- Pass => Crash css/css-overflow/dynamic-visible-to-clip-002.html
! Fail => Crash css/css-overflow/overflow-padding.html
! Skip => Fail css/css-overscroll-behavior/overscroll-behavior-single-axis-keyboard.html
! Skip => Fail css/css-overscroll-behavior/overscroll-behavior-single-axis.html
- Pass => Crash css/css-position/hypothetical-dynamic-change-001.html
- Pass => Crash css/css-position/hypothetical-dynamic-change-002.html
! Fail => Crash css/css-position/hypothetical-dynamic-change-003.html
- Pass => Fail css/css-position/position-absolute-crash-chrome-006.html
- Pass => Crash css/css-text/text-align/text-align-match-parent-03.html
- Pass => Fail css/css-text/text-align/text-align-match-parent-04.html
! Skip => Fail css/css-transforms/hittest-preserve-3d.html
- Pass => Fail css/css-values/calc-nesting.html
- Pass => Crash css/cssom-view/long_scroll_composited.html
- Pass => Fail css/cssom-view/offsetTopLeftInScrollableParent.html
+ Fail => Pass css/cssom/cssstyledeclaration-properties.html
! Skip => Fail css/selectors/active-display-none-001.html
! Fail => Crash css/selectors/dir-style-01b.html
! Fail => Crash css/selectors/dir-style-03b.html
- Pass => Crash css/selectors/invalidation/any-link-attribute-removal.html
! Fail => Timeout svg/animations/beginevents-1.html
! Fail => Timeout svg/animations/eventbase-non-svg-element.html
! Fail => Timeout svg/animations/scripted/eventbase-after-removal.html
! Fail => Timeout svg/animations/slider-switch.html
! Fail => Timeout svg/linking/scripted/a.rel-noreferrer-policy.html
! Fail => Crash svg/struct/scripted/svg-getIntersectionList-003.svg
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/33605626421).</sub>


jerry4718:
I am not sure what form Boa will take to enhance the ability to describe inheritance hierarchies on the Rust side, but this approach should facilitate future migration.

jerry4718:
~There are also some other things here, like `Event`, `EventTarget` [blitz-boa-demo](https://github.com/jerry4718/blitz-boa-demo); I’m not sure if they are appropriate, but I can introduce them gradually if needed.~

jerry4718:
And I'm also not entirely sure if this counts as being somewhat over-engineered.

jerry4718:
~**Known issue:** `el.addEventListener` and `on<event>` do not yet participate in event dispatch with standard semantics.~

~- **JS-side dispatch** (`el.dispatchEvent(...)`): only the listeners registered via `addEventListener` are fired; the `on{event}` property handler is not consulted at all.~
~- **Rust-driven dispatch** (blitz `DomEvent` → runtime): the `on<event>` handler is collected separately and always invoked after every `addEventListener` listener, so the firing order is identical regardless of assignment timing and never assignment-order-sensitive.~
~- The two paths behave inconsistently for the same event object. The `on<event>` properties are currently plain data properties on `Node.prototype` with no registration side effect on assignment, leaving the semantics without a carrier.~

~**Plan:** Turn `on<event>` into accessor properties that register a replaceable listener entry at assignment time and read the current property value at dispatch time, and unify both dispatch paths onto a single listener list.~

jerry4718:
~**Known issue:** Boa does not yet offer a public weak-reference API, which means every Rust-side structure holding a `JsObject` must hold it strongly — there is no way to decouple a reference's lifetime from the GC heap.~

~- **Wrapper cache** (`RuntimeState::node_wrappers`): once a node has been touched by script, its wrapper is held strongly forever. Even after the node is removed from the document (`removeChild`, `innerHTML` rewrites, template re-renders), the wrapper — together with its expando properties and captured closures — has no reclamation path. In long-lived SPA sessions, memory grows monotonically with the number of nodes script has ever touched, and no cleanup strategy can fix this without breaking `===` identity semantics.~
~- **Listener storage** (`EventTargetLayer` own block): the listener list holds callbacks strongly; when the target object dies, its listeners keep the callbacks alive, and there is no way to deregister them when the target is collected.~
~- **Event objects** (`DispatchTarget::Direct` / lazy target cache): an `event.currentTarget` retained by script keeps the (possibly detached) node's wrapper, and with it its listener list, alive.~

~**Plan:** What the cache needs is a **strong/weak switching reference**: one reference per node whose keep-alive strength follows the node's document membership.~

~- **Strong = in document.** An attached node's wrapper carries Rust-side state (registered listeners), so it must stay alive. Initial mode at first wrap: strong iff the node is in the document (Document node exempt, always strong); freshly created or cloned detached nodes wrap weak.~
~- **Weak = detached.** Demote the subtree *before* removal (`removeChild` / `remove` / `replaceWith` / `innerHTML`-detach) — the in-document test walks the parent chain, so demoting after the fact is a no-op and the entry leaks strong forever. Promote on insertion completion (`appendChild` / `insertBefore` / `replaceWith`) with the same in-document gate, recursively over the subtree, so inserting into another detached tree stays weak.~
~- **Entry invalidation.** When a weak entry dies, drop the cache record; re-insertion goes through the promotion path and re-wraps.~

~This can only proceed(on Rust-side) after Boa provides a weak reference API.~
Pinned Boa to 10ef23505ecae734a4984d7839fec99a7ca8482c

jerry4718:
Pinned Boa to 10ef23505ecae734a4984d7839fec99a7ca8482c

PR #823, opened 2026-09-01
Remove hidden nodes from the acccessibility tree

hovinen:
This is as per https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion. It also facilitates testing: in cases where there is only one non-hidden element of a particular role, one can just query for that role without having to specify the query more narrowly.

This includes unit tests of the functionality. To facilitate the assertions it adds a dependency on the crate [Test That!](https://crates.io/crates/test-that).

## WPT results

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/33963784531).</sub>


packages/blitz-dom/src/accessibility.rs
```diff
@@ -32,6 +36,24 @@ impl BaseDocument {
...
+        node.try_stylo_element_data()
+            .as_ref()
+            .and_then(|s| s.get())
+            .map(|s| {
+                s.styles.is_display_none()
+                    || s.styles.primary().clone_visibility()
+                        == visibility::computed_value::T::Hidden
+            })
+            .unwrap_or(false)
+    }
```
nicoburns: Should this not be a method on the node rather than the document if it only uses data from the node?

packages/blitz-dom/src/accessibility.rs
```diff
@@ -222,3 +249,209 @@ fn role_from_element_data(element_data: &ElementData) -> Option<Role> {
         _ => None,
     }
 }
+
+#[cfg(test)]
```
nicoburns: Can you move the tests to the `blitz-tests` crate please?

packages/blitz-dom/src/accessibility.rs
```diff
@@ -222,3 +249,209 @@ fn role_from_element_data(element_data: &ElementData) -> Option<Role> {
...
+
+        verify_that!(
+            tree_update.nodes,
+            contains((
+                anything(),
+                matches_pattern!(AccessKitNode {
+                    role(): eq(Role::GenericContainer)
+                })
+            ))
+        )
```
nicoburns: Not sure about the use of the library here. It introduces an extra DSL that needs to be learnt, and it doesn't seem like it'd be much shorter than a pure-rust assertion...

nicoburns (changes requested):
Core change looks good, but some notes.

packages/blitz-dom/src/accessibility.rs
hovinen: Good point, hadn't considered that. Done.

packages/blitz-dom/src/accessibility.rs
hovinen: Done.

packages/blitz-dom/src/accessibility.rs
hovinen: Of course it's your call. If you want, I will replace it with a built-in assertion.

I found it easier to write the assertions this way. I can write the precise assertion I want in the test without over- or underspecifying, and I get a context-rich failure message for free.

There's an example showing why I care about that in another test which I had to investigate and fix as part of this change: `a_semantic_page_has_no_unknown_elements`. The assertion looks overspecified compared to the name of the test -- if I take the test name literally, it should be asserting that the elements with unknown `Role` is a _subset_ of the given set, not a sequence equal to it. But it's not clear whether this is intentional -- is it important that there be a `head` element with unknown `Role` in the accessibility tree? Did I break something? (For now, I've assumed that the test name is correct, since that seems to make sense, but I'm really not sure.)

nicoburns (approved):

PR #898, opened 2026-09-14
blitz-dom: let callers supply the SVG font database

### LINKED ISSUE #897: SVG text renders blank wherever system fonts are unavailable (e.g. WASM); DocumentConfig.font_ctx does not reach it
# SVG text renders blank wherever system fonts are unavailable (e.g. WASM); `DocumentConfig.font_ctx` does not reach it

## Summary

`<svg><text>` renders nothing in any environment without system fonts. The
documented WASM setup is exactly such an environment.

Blitz has two independent font paths:

| path | used by | configurable via |
|---|---|---|
| `parley::FontContext` | HTML/CSS text | `DocumentConfig.font_ctx` ✅ |
| `usvg::fontdb` | SVG `<text>` | nothing ❌ |

`blitz-dom/src/util.rs`:

```rust
pub(crate) static FONT_DB: LazyLock<Arc<fontdb::Database>> = LazyLock::new(|| {
    let mut db = fontdb::Database::new();
    db.load_system_fonts();
    ...
});
```

It is a private `LazyLock` whose only source is `load_system_fonts()`, so a
caller cannot supply faces to it. `build_single_font_ctx` — whose doc comment
reads *"The standard setup for WASM, where browsers don't expose system
fonts"* — populates the Parley context and has no effect here.

`HOWTO_WASM.md` states *"system fonts aren't available in the browser, so wasm
examples bundle a font at compile time."* Bundling a font fixes HTML text and
leaves SVG text blank, with no error or warning.

## Reproduction

blitz-dom `0.3.0-beta.2`, also reproduced on `main` @ `450e576dd`. Renderer is
blitz-html + blitz-paint + `anyrender_vello_cpu`, with a `font_ctx` built by
`build_single_font_ctx` from three embedded Arial faces.

```html
<!doctype html><meta charset=utf-8>
<body style="margin:0;background:#fff;font:16px Arial">
  <p style="margin:8px">HTML text — uses DocumentConfig.font_ctx</p>
  <svg width="420" height="70" xmlns="http://www.w3.org/2000/svg">
    <text x="8" y="30" font-family="Arial" font-size="22" fill="#000">SVG text</text>
  </svg>
</body>
```

Same binary, two targets, 420x130:

| target | HTML text | SVG text | ink pixels |
| --- | --- | --- | --- |
| host (`aarch64-apple-darwin`) | renders | renders | 5183 |
| `wasm32-wasip1` under wasmtime | renders | **blank** | 1881 |

Isolating just the SVG: native 2117 ink px, wasm **0**.

## Why it also shows up in WPT

Running the WPT runner on both targets and diffing the reports per test
(24,581 tests, 110,757 subtests) gives **zero subtest-level differences** and 17
test-level ones. All 17 are SVG-text tests, and all go native-`FAIL` →
wasm-`PASS` — the wasm side renders no text, which happens to match the
reference. So the failure is currently masked as a pass.

Examples: `css/css-pseudo/svg-text-selection-002.html`,
`css/css-masking/clip-path-svg-content/clip-path-text-001.svg`,
`css/css-text-decor/text-shadow/svg-stroke.html`.

## Suggested fix

Let callers supply the SVG font database — e.g. a `DocumentConfig.svg_font_db`,
or derive it from the faces already registered in `font_ctx` so a single
`build_single_font_ctx` call covers both paths. The latter seems closer to what
users expect, given that helper's stated purpose.

I have this working locally and can open a PR. The shape I tested:
`FONT_DB` becomes a system-font fallback behind an optional supplied database,
plus a public `register_svg_fonts(&[&[u8]])` which also sets the generic
families so `font-family: sans-serif` resolves. One call site changes.

Verified on `wasm32-wasip1`: SVG ink pixels go 0 -> matching the native render
on the repro above. Happy to reshape it if you'd rather thread the faces through
`DocumentConfig` instead.

123Satyajeet123:
`FONT_DB` was only ever filled by `load_system_fonts()`, so SVG `<text>` comes out blank anywhere those don't exist. WASM is the obvious case — and `HOWTO_WASM.md` tells you to bundle a font, which fixes HTML text but not SVG, because `DocumentConfig.font_ctx` only reaches Parley.

This adds `register_svg_fonts(&[&[u8]])` and makes the system database a fallback behind it. It also sets the generic families to the supplied face so `font-family: sans-serif` resolves.

Checked on `wasm32-wasip1`: SVG ink pixels go from 0 to matching the native render. Repro and before/after screenshots are in #897.

Happy to thread the faces through `DocumentConfig` instead if you'd rather keep it off a global.

Fixes #897

## WPT results

Subtests: **2** newly passing, **26** newly failing (net -24).

<details>
<summary>Full diff (28 changed tests)</summary>

```diff
- PASS => FAIL  [0/1]  -1  svg/path/distance/pathLength-positive.svg
- PASS => FAIL  [0/1]  -1  svg/path/distance/pathLength-zero-percentage.svg
- PASS => FAIL  [0/1]  -1  svg/path/distance/pathLength-zero.svg
+ FAIL => PASS  [1/1]  +1  svg/text/reftests/dominant-baseline-text-after-edge.svg
+ FAIL => PASS  [1/1]  +1  svg/text/reftests/dominant-baseline-text-before-edge.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/first-letter.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-bidi-controls-anchors-2.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-complex-001.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-complex-002.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-001.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-002.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-003.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-005.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-006.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-007.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-101.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-inline-size-201.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-multiline-001.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-multiline-002.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-multiline-003.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-shape-inside-001.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/text-shape-inside-002.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/textLength-list-y-spacing.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/textpath-letter-spacing-01.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/textpath-pathlength-css-display-none.tentative.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/textpath-side-001.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/textpath-side-003.svg
- PASS => FAIL  [0/1]  -1  svg/text/reftests/textpath-side-005.svg
```

</details>

<sub>Generated by the [WPT workflow](https://github.com/DioxusLabs/blitz/actions/runs/34904868201).</sub>


nicoburns (changes requested):
Could we use https://docs.rs/usvg/latest/usvg/struct.FontResolver.html to enable SVGs to use Fontique's collection and fallback the same as regular text?

123Satyajeet123:
Reworked to use `FontResolver` as suggested. usvg's `fontdb` now starts empty and the resolver pulls faces from the document's Fontique collection on demand, so `DocumentConfig::font_ctx` is the only font knob and SVG text gets the same families and fallback as HTML text.

`ImageHandler` and `parse_svg_image` take the document's font context to reach it.

Checked on the repro from #897, with `font_ctx` as the only font configuration and no system fonts involved:

| | SVG ink pixels |
| --- | --- |
| before | 0 |
| after, native | 5183 |
| after, `wasm32-wasip1` | 5183 |

Fallback selection walks the generic families and takes the first face whose charmap covers the character. Happy to point it at Fontique's script-based fallback instead if you'd rather.

nicoburns:
> Fallback selection walks the generic families and takes the first face whose charmap covers the character. Happy to point it at Fontique's script-based fallback instead if you'd rather.

Yeah, I reckon this should probably use Fontique's built-in fallback. We should also look into the test regressions.

PR #693, opened 2026-08-12
Add first-party inline SVG rendering (svg-native)

rit3sh-x:
Addresses DioxusLabs/blitz#448

* Adds a `svg-native` feature that parses inline `<svg>` elements directly into Blitz's DOM/Stylo pipeline.
* Inline SVG elements are treated as real, styleable DOM nodes, so cascading, inheritance, `:hover`, and author stylesheets all apply.

## WPT results

Subtests: **19** newly passing, **2** newly failing (net +17). Timeouts: **+3**.

<details>
<summary>Full diff (18 changed tests)</summary>

```diff
- PASS => FAIL       [0/1]  -1  css/compositing/mix-blend-mode/mix-blend-mode-svg.html
+ FAIL => PASS       [1/1]  +1  css/css-flexbox/svg-root-as-flex-item-002.html
+ FAIL => PASS       [1/1]  +1  css/selectors/old-tests/css3-modsel-173a.xml
+ FAIL => PASS       [1/1]  +1  css/selectors/old-tests/css3-modsel-173b.xml
+ FAIL => PASS       [1/1]  +1  css/selectors/old-tests/css3-modsel-174a.xml
+ FAIL => PASS       [1/1]  +1  css/selectors/old-tests/css3-modsel-92.xml
+ FAIL => PASS       [1/1]  +1  css/selectors/old-tests/css3-modsel-93.xml
! FAIL => TIMEOUT    [0/1]  +0  svg/animations/beginelement-instance-time-1.html
+ FAIL => PASS       [2/2]  +1  svg/animations/historical.html
! FAIL => TIMEOUT    [0/4]  +0  svg/animations/scripted/animatetransform-type-missing-value-default.html
! FAIL => TIMEOUT    [0/1]  +0  svg/animations/scripted/keypoints-attribute-trailing-semi.html
- FAIL => FAIL       [1/4]  -1  svg/extensibility/foreignObject/properties.svg
+ FAIL => FAIL     [18/34]  +3  svg/geometry/parsing/height-computed.svg
+ FAIL => FAIL     [18/34]  +3  svg/geometry/parsing/width-computed.svg
+ FAIL => FAIL     [23/39]  +2  svg/interact/scripted/tabindex-focus-flag.svg
+ FAIL => PASS       [1/1]  +1  svg/render/foreignObject-in-non-rendered-getComputedStyle.html
+ FAIL => FAIL       [7/8]  +2  svg/styling/css-selectors-case-sensitivity.html
[...]

PR #904, opened 2026-09-16
svg-native: foundations

rit3sh-x:
## Summary

* Added the `svg-native` feature flag to `blitz-dom` and `blitz-paint`. It is additive and disabled by default.
* Root `<svg>` elements are recognized as `SpecialElementData::SvgRoot` and built into an `SvgContext` after layout, using the content-box as the viewport.
* SVG presentation attributes are now incorporated into the cascade at the `PresHints` level, while author CSS continues to take precedence.
* Added the `CONSTRUCT_SVG` damage bit and an SVG registry, allowing mutations to a single SVG fragment without affecting unrelated fragments or triggering full HTML box reconstruction.

Geometry/painting for shapes, `<use>`, hit-testing, and text are intentionally out of scope for this PR and will be addressed in follow-up PRs.

## Reviewer Note

This PR is a breakdown of #693, which was a very extensive PR. The goal is to split the work into smaller, focused, and independently reviewable chunks.


rit3sh-x:
Hey @nicoburns, I decided to convert #693 to draft and break it down into smaller, more focused, reviewable chunks. I think this will make the changes easier to review and lead to better discussions.

nicoburns:
This seems like a good idea. This is much more reviewable than the previous PR! But perhaps we ought to start with a design document?

rit3sh-x:
Hey @nicoburns, I put together a basic design that I used as a guide while working through the PR:

https://gist.github.com/rit3sh-x/c0aa5f7707009565af31d8f6590fd236

Is it sufficient? I can keep it as is. Otherwise, I can put together a more detailed design.

